PR #7, opened 2025-09-23
Add token permissions to main CI for TestPyPi

anuraaga:


.github/workflows/ci.yaml
```diff
@@ -101,14 +104,6 @@ jobs:
 
       - run: uv sync --frozen
 
-      - name: Modify package name for TestPyPI
```
anuraaga, line 104: I removed this since it seems we may be able to use the correct package name thanks to @mattrobenolt's help

PR #11, opened 2025-09-24
Update getting started doc to follow connectrpc.com patterns

anuraaga:
I mostly followed along the connect-go getting started, picking up a few points about remote plugins from connect-kotlin.

Also fixes some links and adds missing `repr` for headers so they're printable.


stefanvanburen:
> Yeah I think the goal is to get the docs into that repo. I'm thinking of going through page by page to flesh them out here aiming for the doc coverage of other languages and then moving them over. Does that sound ok?

yeah, works for me. I just didn't know if you wanted to just start upstreaming them one-by-one instead as I think docusaurus may have slightly different syntax, but I'm also fine if you want to land them all here and then migrate in one shot :)

PR #15, opened 2025-09-26
Use lowest-direct resolution by default

### LINKED ISSUE #13: Run tests with lowest-direct resolution
Currently, we don't have `uv`'s dependency resolution configured, so although we declare lower bounds on (at least our production) dependencies, we're only testing with [`resolution=highest`](https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy). Ideally, we'd add a matrix dimension where we tested with [resolution=lowest](https://docs.astral.sh/uv/reference/cli/#uv-run) so we know that our lower bounds are valid. From uv's docs:

> When publishing libraries, it is recommended to separately run tests with --resolution lowest or --resolution lowest-direct in continuous integration to ensure compatibility with the declared lower bounds.

This will _probably_ result in us adding a lower bound to [`httpx`](https://github.com/connectrpc/connect-python/blob/f4eb908d9cab43f0a2e6d6ee485c952b97070945/pyproject.toml#L12) (and perhaps our [development dependencies](https://github.com/connectrpc/connect-python/blob/f4eb908d9cab43f0a2e6d6ee485c952b97070945/pyproject.toml#L37-L54)).

Here's an [example PR where I did this in protovalidate-python](https://github.com/bufbuild/protovalidate-python/pull/355).

If we want to do this, I think it makes sense to configure `uv` to use `lowest-direct` by default locally (again, [protovalidate-python PR example](https://github.com/bufbuild/protovalidate-python/pull/373)), and then iterate on the dependency bounds until we get the checks working locally.

stefanvanburen:
WIP. Trying to keep our production dependencies as wide as is reasonable, but fine to pin our development dependencies tight. Dropped the gencode down to the plugin version that works with protobuf==5 (from protovalidate-python). Worked around a couple of minor issues.

Fixes #13.


pyproject.toml
```diff
@@ -35,28 +35,32 @@ Issues = "https://github.com/connectrpc/connect-python/issues"
...
+    "hypercorn>=0.17.3",
+    "just-bin>=1.42.4; sys_platform != 'win32'",
+    "mkdocs>=1.6.1",
+    "mkdocs-material>=9.6.20",
+    "mkdocstrings[python]>=0.30.1",
+    "pyright[nodejs]>=1.1.405",
+    "pytest>=8.4.2",
+    "pytest-asyncio>=1.1.0",
+    "pytest-cov>=7.0.0",
+    "ruff~=0.12.12", # Stay compatible with 0.12 until we want to bump to 0.13, where formatting may change
```
anuraaga: I wonder if we should use strict versions for all the dev dependencies (maybe not the optional runtime ones like brotli, zstandard)? I wouldn't want someone (especially an external contributor) to have their PR fail because `highest` ran with some newer version of pytest that broke something for example - we basically lose a lot of the reason for having a lock file.

Ideally we could leave that to CI running on an auto-update workflow but I'm not sure dependabot handles pyproject.toml updates for uv projects, renovate would.

I think we can follow up on this though instead of requiring for this PR

pyproject.toml
```diff
@@ -9,7 +9,7 @@ maintainers = [
     { name = "Yasushi Itoh", email = "i2y.may.roku@gmail.com" },
 ]
 requires-python = ">= 3.10"
-dependencies = ["httpx", "protobuf>=5.28"]
+dependencies = ["httpx>=0.28.1", "protobuf>=5.28"]
```
anuraaga: Thanks for finding the floor probably was pretty annoying :)

example/pyproject.toml
```diff
@@ -2,15 +2,15 @@
...
-[tool.ruff]
-extend = "../pyproject.toml"
+dependencies = ["connect-python", "flask>=3.1.2", "protobuf>=5.28", "starlette>=0.48.0"]
 
 [build-system]
 requires = ["uv_build>=0.7.21,<0.8.0"]
 build-backend = "uv_build"
 
+[tool.ruff]
+extend = "../pyproject.toml"
```
anuraaga, line 16: Thanks always forget this

test/buf.gen.yaml
```diff
@@ -1,8 +1,9 @@
 version: v2
 plugins:
-  - remote: buf.build/protocolbuffers/python:v32.0
+  # NOTE: v26.0 is the earliest version supporting protobuf==5.
```
anuraaga, line 3: IIRC, this causes a warning when running with the latest protobuf, which would happen on CI for `highest`, and eventually it might fail completely if such compatibility ends. 

This is probably fine, especially since I can't think of anyway to improve it... Maybe we should copy the above content into the comment?

pyproject.toml
```diff
@@ -215,7 +219,7 @@ exclude = [
 ]
 
 [tool.uv.workspace]
-members = ["conformance", "example", "noextras"]
```
anuraaga, line 218: I think the main reason I had it as a workspace member was I didn't think pytest would behave well otherwise, but apparently it does.

I also have a longish term idea of running tests with more app servers, HTTP/1 uvicorn, WSGI gunicorn, etc. It's probably fine to add them all in dev-dependencies here but what do you think?

pyproject.toml
stefanvanburen: hah - I was going to comment that this is almost certainly _not_ the floor, actually just the latest release, but was going to see if we could _start_ with being strict about this dep and loosen it if someone asks (not a breaking change).

I think with httpx being pre-1.0 and this release being 9 months old, I imagine most projects using it are probably pretty close to the bleeding edge of releases. Happy to try to loosen it here, though.

pyproject.toml
stefanvanburen: yeah, to be honest this is my first experience w/ uv workspaces so I wasn't sure how best to use them; with adding all the dependency specifiers, I thought it would be nice if we just consolidated on basically two `pyproject.toml`s for the implementation and the plugin. I agree that I think it would be best if we just had them all in the top-level dev dependencies of pyproject.toml.

Not going to try to remove the other two workspace members in this PR, but something to think about?

test/buf.gen.yaml
stefanvanburen: I _think_ the newest versions of `protobuf` removed a lot of those warnings as they were noisy; this is pretty much the same setup as `protovalidate-python` and while it [has some warnings in tests from protobuf when running on `highest`](https://github.com/bufbuild/protovalidate-python/actions/runs/18025204602/job/51290853781#step:7:47), I don't think we'll hit them in connect-python.

Separately, I just noticed that [they added a section to the compatibility page on protobuf.dev for python](https://protobuf.dev/support/cross-version-runtime-guarantee/#python) — basically, my read of that is that in theory, someone running `protobuf<8` could use gencode all the way back to the 3.20 plugin release. I think this is a reasonable middle ground, though.

pyproject.toml
stefanvanburen: > if we should use strict versions for all the dev dependencies

do you mean just pin the versions for dev dependencies? I'd be happy with that.

> Ideally we could leave that to CI running on an auto-update workflow but I'm not sure dependabot handles pyproject.toml updates for uv projects, renovate would.

yeah, dependabot isn't great for `uv` yet :(

> I wouldn't want someone (especially an external contributor) to have their PR fail because `highest` ran with some newer version of pytest that broke something for example - we basically lose a lot of the reason for having a lock file.

👍

example/pyproject.toml
stefanvanburen: it just moved down in the diff, don't worry, you didn't forget :).

my linter was just complaining about the toml ordering of keys; I'll leave this change but not bother with it in the future.

stefanvanburen:
Ready for looks; open to thoughts on the approach and otherwise!

test/buf.gen.yaml
anuraaga: Oh nice - I didn't expect protobuf team to ever provide such wide compatibility guarantees ;) But then this looks even better

pyproject.toml
anuraaga: I think conformance was definitely the least compelling usage of workspaces as we see here. But the others have more technical reasons

- example as a built project becomes trivial to import for "testable examples" while still being separate enough to be an example. I believe accomplishing it without the project would require pythonpath manipulation which can get tricky

- noextras needs to exclude the optional runtime dependencies, and uv provides an easy way to do that with `--exact` on the smaller pyproject.toml

Both can probably be worked around if we need to. IIUC the biggest problem with multiple projects is having to propagate version pins across all the projects which is indeed tedious. uv provides dependency management at the top level with for example

```
[tool.uv]
override-dependencies = ["pytest ==8.42"]
```

Would this be a reasonable approach to keep the above? Maybe it's too weird though

pyproject.toml
anuraaga: Ah oops - got it. Yeah that strategy makes sense to me

pyproject.toml
anuraaga: > do you mean just pin the versions for dev dependencies? I'd be happy with that.

Yeah that's what I was thinking. BTW out of curiosity any thoughts on using renovate? I guess it may not make sense to run on a single repo in the connectrpc org but not sure.

stefanvanburen: I think it'd be ok if we switch to renovate; I have heard that it's a lot better for `uv` at this point, especially if it helps out with syncing the version pins across the pyproject.tomls. (I think we're also the only connectrpc project using codecov and that's fine as well.)

And cool, I can just update these to direct pins for now and see if renovate can bump both the pins and uv.lock for dev dependencies?

pyproject.toml
stefanvanburen: I'm fine with keeping them around if there's a good technical reason, honestly I just didn't grok all of it at first :). Might be less of a big deal if renovate helps with keeping the pins in sync.

pyproject.toml
stefanvanburen: Pinned the dev deps in https://github.com/connectrpc/connect-python/pull/15/commits/b8b5cf9da2b7fd3542cf7d26764142a63ab58fbf.

pyproject.toml
```diff
@@ -2,220 +2,220 @@
...
-    "Operating System :: OS Independent",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.10",
-    "Programming Language :: Python :: 3.11",
-    "Programming Language :: Python :: 3.12",
-    "Programming Language :: Python :: 3.13",
-    "Topic :: Internet :: WWW/HTTP",
-    "Topic :: Software Development :: Libraries :: Python Modules",
-    "Typing :: Typed",
+  "Development Status :: 3 - Alpha",
```
anuraaga: Just a double check, FWICT the indent change is from bumping ruff to 0.13. Not a problem, I get the same formatting when checking out the PR and just wanted to make sure it's not an environment specific change.

stefanvanburen: ugh, no, not a `ruff` change. sorry, have [tombi](https://github.com/tombi-toml/tombi) enabled locally and didn't even notice it was changing things 🤦 . Let me revert that to not confuse things.

anuraaga: Interesting - I guess my attempts to break the formatting must have been thwarted by some heuristics on the indents within the files.

I'm a fan of auto formatting so let's follow up on whether we can add tombi to the CI

stefanvanburen: sounds good. I know tombi is pre-1.0 but it's been pretty good so far. I've also seen projects like https://github.com/tox-dev/toml-fmt that are more pyproject-specific that we could look at. Open to whatever; agree that auto-formatting is nice.

PR #16, opened 2025-09-29
Add doc for headers and trailers

anuraaga:
Follows pattern of connect-go docs fairly closely. Will follow up with binary headers after also adding the encoding functions, which are particularly important in Python since it's not easy to get unpadded base64 in python.


i2y (approved):
✨

stefanvanburen (approved):

PR #17, opened 2025-09-30
Add doc on streaming

anuraaga:
Similar to Go docs but omits headers and interceptors sections since here there is no difference for headers based on stream type, and the streaming interceptor payload interceptors are document in `interceptors.md`.


docs/streaming.md
```diff
@@ -0,0 +1,133 @@
...
+Let's start by amending the `GreetService` we defined in [Getting Started](./getting-started.md) to make the Greet method use
+client streaming:
+
+```protobuf
+syntax = "proto3";
+
+package greet.v1;
+
+option go_package = "example/gen/greet/v1;greetv1";
+
```
stefanvanburen:
```suggestion
```

anuraaga: Definitely, thanks for noticing

PR #19, opened 2025-10-03
Add doc on error handling

i2y:


docs/errors.md
```diff
@@ -0,0 +1,192 @@
...
+  "code": "invalid_argument",
+  "message": "name is required",
+  "details": [
+    {
+      "type": "google.protobuf.Struct",
+      "value": "base64-encoded-protobuf"
+    }
+  ]
+}
+```
```
stefanvanburen: nit: not sure if `http` is a valid markdown language tag, but could be more appropriate here if it is. (although then maybe the `json` wouldn't be pretty 😞.)

can probably just ignore this.

docs/usage.md
```diff
@@ -159,3 +159,148 @@ class ElizaServiceImpl:
...
+## Error Handling Best Practices
+
+### Choosing appropriate error codes
+
+Select error codes that accurately reflect the situation:
+
+- Use `INVALID_ARGUMENT` for malformed requests that should never be retried
+- Use `FAILED_PRECONDITION` for requests that might succeed if the system state changes
+- Use `UNAVAILABLE` for transient failures that should be retried
+- Use `INTERNAL` sparingly - it indicates a bug in your code
```
stefanvanburen, line 172: I might link to https://connectrpc.com/docs/protocol#error-codes; the discussion below the table about how to choose an error code is pretty useful.

docs/usage.md
```diff
@@ -159,3 +159,148 @@ class ElizaServiceImpl:
...
+```python
+from google.rpc.error_details_pb2 import BadRequest
+
+# Good - structured details
+bad_request = BadRequest()
+for field, error in validation_errors.items():
+    violation = bad_request.field_violations.add()
+    violation.field = field
+    violation.description = error
+raise ConnectError(Code.INVALID_ARGUMENT, "Validation failed", details=[bad_request])
```
stefanvanburen, line 201: my only hesitation on recommending error details is that in practice, they can make it more difficult to debug if your client isn't already deserializing them. And there's no way to strongly enforce them in your API contract; the best you can really do is just document them.

I think this is fine to leave as-is, just wish error details were easier to use in general.

anuraaga: I think this PR is trying to match the coverage of Go docs which similarly cover error details so is probably good as is. Agree that error details are a bit clunky but too useful for substituting for custom error codes to leave out.

anuraaga (approved):

docs/usage.md
i2y: I see, thanks!  I just wrote the link below.

docs/usage.md
i2y: I see. I agree with both opinions, so I just added this https://github.com/connectrpc/connect-python/pull/19/files#diff-72376d0b487d7231cf31959bf266f6aea098e899743bae02e36d176cef4db476R210

docs/errors.md
i2y: Thanks!  I changed `json` to `http` now

stefanvanburen (approved):

PR #20, opened 2025-10-03
Add doc about gRPC (lack of) compatibility and migration path

anuraaga:


docs/grpc-compatibility.md
```diff
@@ -0,0 +1,98 @@
+# gRPC compatibility
```
anuraaga, line 1: Put some thought into whether this is actually the right title when there isn't compatibility, but it seemed still fine but open to any thoughts

docs/grpc-compatibility.md
```diff
@@ -0,0 +1,98 @@
...
+
+Connect-Python currently does not support the gRPC protocol due to lack of support for HTTP/2 trailers
+in the Python ecosystem. If you have an existing codebase using grpc-python and want to introduce Connect
+in a transition without downtime, you will need a way for the gRPC servers to be accessible from both
+gRPC and Connect clients at the same time. envoy is a widely used proxy server with support for translating
```
stefanvanburen: nit: I see "envoy" stylized both all lowercase and capitalized on https://www.envoyproxy.io/docs/envoy/latest/; I think I'd opt for caps here?

```suggestion
gRPC and Connect clients at the same time. Envoy is a widely used proxy server with support for translating
```

Looking a little more, although the logomark is lowercase, the rest of the docs seem to refer to it as Envoy (and so do [the Connect docs](https://connectrpc.com/docs/faq/#does-connect-rpc-support-envoy)). So I might just `s/envoy/Envoy` throughout this doc?

i2y (approved):

docs/grpc-compatibility.md
anuraaga: Thanks - yeah eyes landed on the only two instances of lowercase in those docs for some reason

PR #23, opened 2025-10-04
Add doc on testing

i2y:
This PR adds comprehensive testing guide with ASGI/WSGI examples for services, clients, and interceptors.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -4,14 +4,7 @@ This guide covers testing connect-python services and clients.
 
 ## Setup
 
-Install httpx for testing connect-python services and clients:
-
-```bash
-pip install httpx
-# or: uv add --dev httpx
-```
-
-For pytest examples in this guide, you'll also need pytest (and pytest-asyncio for async tests). unittest requires no additional dependencies.
+For pytest examples in this guide, you'll need pytest and pytest-asyncio. unittest requires no additional dependencies.
 
 ## Recommended approach: In-memory testing
 


docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+            self.requests = []
+
+        async def on_start(self, ctx):
+            method_name = ctx.method().name
+            self.requests.append(method_name)
+            return method_name
+
+        async def on_end(self, token, ctx):
+            # token is the value returned from on_start
+            pass
```
stefanvanburen: do we need to return the `method_name` / have the comment if we just check `self.requests`? up to you if you want to keep this in (similar comment for `WSGI`)

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+
+### In-memory testing
+
+Test services using httpx's ASGI/WSGI transport, which tests your full application stack while remaining fast and isolated:
+
+=== "ASGI"
+
+    ```python
+    import pytest
+    import httpx
```
anuraaga:
```suggestion
    import httpx
    import pytest
```

Extreme nit

docs/testing.md
```diff
@@ -0,0 +1,923 @@
+# Testing Guide
```
anuraaga:
```suggestion
# Testing
```

There are a couple existing docs with this title

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+
+=== "ASGI"
+
+    ```python
+    import pytest
+    import httpx
+    from greet.v1.greet_connect import GreetService, GreetServiceASGIApplication, GreetServiceClient
+    from greet.v1.greet_pb2 import GreetRequest, GreetResponse
+
+    class TestGreetService(GreetService):
```
anuraaga: For the section about testing services, I think we wouldn't have names like `TestService` - we should be testing the real code here.

I think we can do something like 

`Testing the service we created in the [Getting Started](./getting-started) guide may look like this:`

```
from greet_connect import GreetASGIApplication
from server import Greeter
``` 
any real test will be importing the service, not having it defined in the same file

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+        async def greet(self, request, ctx):
+            return GreetResponse(greeting=f"Hello, {request.name}!")
+
+    @pytest.mark.asyncio
+    async def test_greet():
+        # Create the ASGI application
+        app = GreetServiceASGIApplication(TestGreetService())
```
anuraaga: We can't show it clearly with our example services, we could consider something like `# Initialize your service with mock dependencies as needed`. Though maybe we can leave "general testing" advice out of the doc

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+        ) as session:
+            yield GreetServiceClient("http://test", session=session)
+
+    @pytest.mark.asyncio
+    async def test_greet(greet_client):
+        response = await greet_client.greet(GreetRequest(name="Alice"))
+        assert response.greeting == "Hello, Alice!"
+
+    @pytest.mark.asyncio
+    async def test_greet_multiple_names(greet_client):
```
anuraaga: There doesn't seem to be multiple names. How about `empty_name` and use empty string?

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+            client = GreetServiceClientSync("http://test", session=session)
+
+            with pytest.raises(ConnectError) as exc_info:
+                client.greet(GreetRequest(name=""))
+
+        assert exc_info.value.code == Code.INVALID_ARGUMENT
+        assert "name is required" in exc_info.value.message
+    ```
+
+### Testing streaming services
```
anuraaga: I don't think there's anything special about testing streaming services, these are more examples on how to use streaming. I think there is more of this we can add to `streaming.md` does it make sense to move them there instead of including in this file?

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+                yield GreetRequest(name="Alice")
+                yield GreetRequest(name="Bob")
+
+            response = client.greet_many(request_stream())
+
+        assert "Alice" in response.greeting
+        assert "Bob" in response.greeting
+    ```
+
+### Testing with context (headers and trailers)
```
anuraaga: Maybe this one overlaps too much with the headers doc. We could add a sentence above that covers it, something like `With your test client setup, you can use any connect code for interacting with the service under test including streaming, reading headers or checking errors.` and link to those docs? If we miss any coverage, we can add to those docs instead of here.

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+    ```
+
+This pattern:
+
+- Reduces code duplication across multiple tests
+- Makes tests more readable and focused on behavior
+- Follows pytest best practices
+- Matches the pattern used in connect-python's own test suite
+
+### Testing error handling
```
anuraaga: After writing below comments, I also thought this overlaps well with errors.md.

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+        except ConnectError as e:
+            if e.code == Code.NOT_FOUND:
+                return "User not found"
+            elif e.code == Code.UNAUTHENTICATED:
+                return "Please login"
+            raise
+
+    @pytest.mark.asyncio
+    async def test_client_error_handling():
+        class TestGreetService(GreetService):
```
anuraaga, line 284: Is it worth something like `# It can be convenient to use a mock such as with pytest mocker`?

docs/testing.md
anuraaga: Agree that along the lines of above comments, it's best if we show how to test something real since users will be reading this for advice on that. This `LoggingInterceptor` doesn't actually log - to make it a testable real interceptor, we'd probably pass in a `Logger` and mock it or similar. Alternatively, auth interceptor may be easier to demonstrate here

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+
+        # Verify interceptor was called
+        assert "Greet" in interceptor.requests
+    ```
+
+## Test organization
+
+### Project structure
+
+Organize your tests in a `test/` directory at the root of your project:
```
anuraaga: Actually this is something I forgot when reorganizing for this repo - AFAIK, `tests` is the default for pytest and doesn't require configuration, that's what we should use here.

That being said, this whole section seems more like docs on how to use pytest, rather than testing connect. Maybe we can remove it?

docs/testing.md
```diff
@@ -0,0 +1,923 @@
...
+    from connectrpc.code import Code
+    from connectrpc.errors import ConnectError
+    from greet.v1.greet_connect import GreetService, GreetServiceASGIApplication, GreetServiceClient
+    from greet.v1.greet_pb2 import GreetRequest, GreetResponse
+
+    class AuthGreetService(GreetService):
+        async def greet(self, request, ctx):
+            # Check for authorization header
+            auth = ctx.request_headers().get("authorization")
+            if not auth or not auth.startswith("Bearer "):
```
anuraaga: If we update the interceptor example, we can remove this one, which is good since I think it's very far from any real life service

i2y:
@anuraaga @stefanvanburen 
Thanks! I've addressed your comments. Could you please review this PR again?

docs/testing.md
```diff
@@ -0,0 +1,460 @@
+# Testing
+
+This guide covers testing connect-python services and clients.
+
+## Setup
+
+Install httpx for testing connect-python services and clients:
```
anuraaga: I don't know if we need this, won't it always be installed as a dependency of connect-python?

anuraaga (approved):
Looks much nicer, thanks!

docs/testing.md
i2y: Thanks for catching that! Fixed in 58d6653.

stefanvanburen (approved):

PR #29, opened 2025-10-08
Add doc for client metadata interceptor

### LINKED ISSUE #28: Provide an example of setting client auth headers
The documentation talks about how to check auth headers in the service: https://github.com/connectrpc/connect-python/blob/main/docs/interceptors.md#metadata-interceptors
As well as how to set headers from a client on a per request basis, but does not show how to add a header to all requests using an interceptor. Adding some documentation for this would be great!

Thanks.

anuraaga:
#28 gave the idea of fleshing out the auth interceptors to propagate the auth token from server to client. Since the two interceptors demonstrate the two forms of on_end or not, I removed the timing interceptor.

Fixes #28

/cc @rockwotj thanks for the idea!

### DIFF
--- a/docs/interceptors.md
+++ b/docs/interceptors.md
@@ -193,7 +193,7 @@ An authentication interceptor checking bearer tokens and storing them to a conte
             if not authorization or not authorization.startswith("Bearer "):
                 raise ConnectError(Code.UNAUTHENTICATED)
             token = authorization[len("Bearer "):]
-            if token not in valid_tokens:
+            if token not in self._valid_tokens:
                 raise ConnectError(Code.PERMISSION_DENIED)
             return _auth_token.set(token)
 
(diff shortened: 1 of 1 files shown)


docs/interceptors.md
```diff
@@ -175,65 +175,91 @@ to be able to intercept RPC messages. However, many interceptors, such as for au
...
         def __init__(self, valid_tokens: list[str]):
             self._valid_tokens = valid_tokens
 
-        async def on_start(self, ctx: RequestContext):
+        async def on_start(self, ctx: RequestContext) -> Token["auth_token"]:
             authorization = ctx.request_headers().get("authorization")
             if not authorization or not authorization.startswith("Bearer "):
                 raise ConnectError(Code.UNAUTHENTICATED)
             token = authorization[len("Bearer "):]
             if token not in valid_tokens:
```
i2y:
```suggestion
            if token not in self._valid_tokens:
```

docs/interceptors.md
```diff
@@ -175,65 +175,91 @@ to be able to intercept RPC messages. However, many interceptors, such as for au
...
         def __init__(self, valid_tokens: list[str]):
             self._valid_tokens = valid_tokens
 
-        def on_start(self, ctx: RequestContext):
+        def on_start(self, ctx: RequestContext) -> Token["auth_token"]:
             authorization = ctx.request_headers().get("authorization")
             if not authorization or not authorization.startswith("Bearer "):
                 raise ConnectError(Code.UNAUTHENTICATED)
             token = authorization[len("Bearer "):]
             if token not in valid_tokens:
```
i2y:
```suggestion
            if token not in self._valid_tokens:
```

PR #37, opened 2025-10-21
Do multiple reads for WSGI even with content length

anuraaga:
For `.read()` methods

https://docs.python.org/3/library/io.html#io.RawIOBase.read

> Fewer than size bytes may be returned if the operating system call returns fewer than size bytes.

In practice, all WSGI servers seem to buffer the request to content-length before calling the app...except for [pyvoy](https://github.com/curioswitch/pyvoy) ;) Incidentally that means pyvoy is able to pass bidi streaming tests with WSGI too. I'm currently debugging an occasional macOS crash which seems envoy-related, but am feeling pretty good about the project.

I also took care to follow the WSGI spec "The application should not attempt to read more data than is specified by the CONTENT_LENGTH variable.". If this text wasn't there, it would be easiest to just call `.read()` with no parameter.

https://peps.python.org/pep-3333/#input-and-error-streams

I'm interested in adding pyvoy to conformance tests after some more debugging, but without it it's quite challenging to add a regression test for this change. I wanted to separate out the logic change from any testing changes but let me know if I should hold off until having that regression test.

### DIFF
--- a/src/connectrpc/_server_sync.py
+++ b/src/connectrpc/_server_sync.py
@@ -118,9 +118,16 @@ def prepare_response_headers(
 def _read_body_with_content_length(
     environ: WSGIEnvironment, content_length: int
 ) -> bytes:
-    bytes_read = 0
-    chunks = []
     input_stream: BytesIO = environ["wsgi.input"]
+
+    # Many app servers buffer the entire request before executing the app
+    # so do an optimistic read before looping.
+    chunk = input_stream.read(content_length)
+    if len(chunk) == content_length:
+        return chunk
+
+    bytes_read = len(chunk)
+    chunks = [chunk]
     while bytes_read < content_length:
         to_read = content_length - bytes_read
         chunk = input_stream.read(to_read)
@@ -128,6 +135,11 @@ def _read_body_with_content_length(
             break
         chunks.append(chunk)
         bytes_read += len(chunk)
+    if bytes_read < content_length:
+        raise ConnectError(
+            Code.INVALID_ARGUMENT,
+            f"request truncated, expected {content_length} bytes but only received {bytes_read} bytes",
+        )
     return b"".join(chunks)
 
 


src/connectrpc/_server_sync.py
```diff
@@ -115,6 +115,22 @@ def prepare_response_headers(
     return headers
 
 
+def _read_body_with_content_length(
+    environ: WSGIEnvironment, content_length: int
+) -> bytes:
+    bytes_read = 0
+    chunks = []
```
stefanvanburen: nit: could we pre-allocate a bytearray to be size `content_length` and then use [`readinto`](https://docs.python.org/3/library/io.html#io.RawIOBase.readinto)?

src/connectrpc/_server_sync.py
```diff
@@ -115,6 +115,22 @@ def prepare_response_headers(
...
+    input_stream: BytesIO = environ["wsgi.input"]
+    while bytes_read < content_length:
+        to_read = content_length - bytes_read
+        chunk = input_stream.read(to_read)
+        if not chunk:
+            break
```
spenczar, line 135: It surprises me that short reads don't raise an exception. I think this could result in odd behavior, and that it would be better to raise if we don't wind up with a body of the expected length.

src/connectrpc/_server_sync.py
anuraaga: WSGI doesn't actually define `readinto`.

https://peps.python.org/pep-3333/#input-and-error-streams

Since most servers will prebuffer anyways, I added a small optimization for that case instead

src/connectrpc/_server_sync.py
anuraaga: Thanks! At first I thought it would be fine as an invalid protobuf but realized that there are truncations that can result in a valid proto. Added an error for it.

spenczar (approved):

PR #40, opened 2025-10-30
Add daphne to conformance test

anuraaga:
Trying to find a full-duplex capable server that reliably passes our tests, gave daphne a try. I was optimistic since it supports websockets, but it turns out that their http protocol is completely separate and doesn't support bidi... Oh well, having added the test figured it's worth running it as an option for the others.

### DIFF
--- a/conformance/test/server.py
+++ b/conformance/test/server.py
@@ -10,7 +10,7 @@ import time
 from collections.abc import AsyncIterator, Iterator
 from contextlib import ExitStack, closing
 from tempfile import NamedTemporaryFile
-from typing import TYPE_CHECKING, Literal, TypeVar
+from typing import TYPE_CHECKING, Literal, TypeVar, get_args
 
 from _util import create_standard_streams
 from gen.connectrpc.conformance.v1.config_pb2 import Code as ConformanceCode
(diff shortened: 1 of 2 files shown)


conformance/test/server.py
```diff
@@ -415,22 +415,64 @@ def _server_env(request: ServerCompatRequest) -> dict[str, str]:
...
 
 async def _tee_to_stderr(stream: asyncio.StreamReader) -> AsyncIterator[bytes]:
-    try:
-        while True:
-            line = await stream.readline()
-            if not line:
-                break
-            print(line.decode("utf-8"), end="", file=sys.stderr)  # noqa: T201
-            yield line
-    except asyncio.CancelledError:
```
anuraaga, line 425: I had accidentally added this before I think. It ends up swallowing the error preventing `terminate` from being called on the subprocesses

conformance/test/server.py
```diff
@@ -619,14 +657,14 @@ def _find_free_port():
 
 class Args(argparse.Namespace):
     mode: Literal["sync", "async"]
-    server: Literal["granian", "hypercorn", "uvicorn"]
+    server: Literal["daphne", "granian", "hypercorn", "uvicorn"]
```
stefanvanburen: nit: should this also have "gunicorn" in the list?

conformance/test/test_server.py
```diff
@@ -87,13 +87,23 @@ def test_server_sync(server: str) -> None:
         pytest.fail(f"\n{result.stdout}\n{result.stderr}")
 
 
-@pytest.mark.parametrize("server", ["granian", "hypercorn", "uvicorn"])
+@pytest.mark.parametrize("server", ["daphne", "granian", "hypercorn", "uvicorn"])
```
stefanvanburen, line 90: should `"daphne"` be added to `test_server_sync`, above?

pyproject.toml
```diff
@@ -53,6 +54,7 @@ dev = [
     "pytest-cov==7.0.0",
     "ruff~=0.13.2",
     "uvicorn==0.37.0",
+    "Twisted[tls,http2]==25.5.0",
```
stefanvanburen, line 58: does `daphne` dep on a lower version of twisted? (oh, ah, I guess it's for http2 support: https://github.com/django/daphne?tab=readme-ov-file#http2-support)

conformance/test/server.py
anuraaga: Yeah I guess there's something causing pyright to not flag this

conformance/test/test_server.py
anuraaga: Daphne doesn't support wsgi so only async.

pyproject.toml
anuraaga: Will add a comment

PR #46, opened 2025-11-11
Add support for ASGI lifespan

anuraaga:
I realized we need to support lifespan events to allow asyncio startup / shutdown logic for an app. The simplest way to expose this to users seemed to be accepting either a service or an async generator of a service in the application and resolve either within the lifespan events. We fallback to resolving on first request if the server doesn't support lifespan (this is basically just applying interceptors and shouldn't be a too-large first request penalty), one that notably doesn't is the httpx mock transport we recommend for unit tests.

Let me know any thoughts including the general approach.

### DIFF
--- a/conformance/test/server.py
+++ b/conformance/test/server.py
@@ -531,7 +531,12 @@ async def serve_gunicorn(
     cafile: str | None,
     port_future: asyncio.Future[int],
 ):
-    args = ["--bind=127.0.0.1:0", "--threads=40", "--reuse-port"]
+    args = [
+        "--bind=127.0.0.1:0",
+        "--workers=4",
+        "--worker-class=gevent",
+        "--reuse-port",
+    ]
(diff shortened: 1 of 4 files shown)


src/connectrpc/_server_async.py
```diff
@@ -64,45 +73,96 @@
 )
 
 
-class ConnectASGIApplication(ABC):
+class ConnectASGIApplication(Generic[_SVC]):
```
i2y: Since this class has abstract methods, maybe it should also inherit from `ABC`?

anuraaga: Right, not sure why I thought Generic would imply that - fixed

anuraaga:
@i2y Can you give this another look? Also @stefanvanburen @spenczar in case this went under the radar. Thanks!

PR #50, opened 2025-11-14
Run conformance test with pyvoy

anuraaga:
As pyvoy was made with connect-python in mind, I guess it's not surprising it seems to handle the conformance tests fairly reliably.

Unfortunately, I haven't been able to get good results from granian and hypercorn so have gone ahead and removed them from the default runs, the wiring is still in place to try them again if things improve in the future.

### DIFF
--- a/conformance/test/test_server.py
+++ b/conformance/test/test_server.py
@@ -46,9 +46,7 @@ def test_server_sync(server: str) -> None:
         [sys.executable, _server_py_path, "--mode", "sync", "--server", server]
     )
     opts = [
-        # While Hypercorn and Granian supports HTTP/2 and WSGI, they both have simple wrappers
-        # that reads the entire request body before running the application, which does not work for
-        # full duplex.
+        # TODO: Enable full-duplex in pyvoy
         "--skip",
         "**/bidi-stream/full-duplex/**",
     ]


conformance/test/test_server.py
```diff
@@ -37,24 +37,20 @@ def macos_raise_ulimit():
...
     args = maybe_patch_args_with_debug(
         [sys.executable, _server_py_path, "--mode", "sync", "--server", server]
     )
     opts = [
-        # While Hypercorn and Granian supports HTTP/2 and WSGI, they both have simple wrappers
-        # that reads the entire request body before running the application, which does not work for
-        # full duplex. There are no other popular WSGI servers that support HTTP/2, so in practice
-        # it cannot be supported. It is possible in theory following hyper-h2's example code in
-        # https://python-hyper.org/projects/hyper-h2/en/stable/wsgi-example.html though.
+        # TODO: Enable full-duplex in pyvoy
```
anuraaga, line 49: This worked with some initial testing but I need to validate it some more

conformance/test/_util.py
```diff
@@ -1,7 +1,7 @@
 import asyncio
 import sys
 
-VERSION_CONFORMANCE = "v1.0.4"
+VERSION_CONFORMANCE = "0bed9ca203aeda4e344edc442a4b2ede91726db5"
```
stefanvanburen, line 4: want to create an issue over in https://github.com/connectrpc/conformance/issues to tag a new version for https://github.com/connectrpc/conformance/pull/1047? cc @jhump @doriable

stefanvanburen (approved):

conformance/test/_util.py
anuraaga: Sure filed https://github.com/connectrpc/conformance/issues/1050

PR #51, opened 2025-11-18
Enable wsgi streaming conformance tests with pyvoy

anuraaga:
And tweaks gunicorn

### DIFF
--- a/conformance/test/server.py
+++ b/conformance/test/server.py
@@ -533,7 +533,7 @@ async def serve_gunicorn(
 ):
     args = [
         "--bind=127.0.0.1:0",
-        "--workers=8",
+        "--workers=4",
         "--worker-class=gevent",
         "--reuse-port",
     ]


conformance/test/server.py
```diff
@@ -531,7 +531,12 @@ async def serve_gunicorn(
     cafile: str | None,
     port_future: asyncio.Future[int],
 ):
-    args = ["--bind=127.0.0.1:0", "--threads=40", "--reuse-port"]
+    args = [
```
anuraaga, line 534: Since updating conformance to have connection reuse, gunicorn would have some flakes. A bit disappointed since the vanguard has been so robust otherwise, but doing some research it seems a lot of settings assume nginx in front of gunicorn.

Through some random selection I got these settings to pass CI 3 times, but if anyone happens to know gunicorn better happy to hear any thoughts.

pyproject.toml
```diff
@@ -53,7 +53,7 @@ dev = [
     "pytest-asyncio==1.2.0",
     "pytest-cov==7.0.0",
     "pytest-timeout==2.4.0",
-    "pyvoy==0.1.1; sys_platform != 'win32'",
+    "pyvoy==0.1.2; sys_platform != 'win32'",
```
anuraaga, line 56: Fixed an issue that could sporadically hang with WSGI

PR #52, opened 2025-11-21
Add doc on deployment

anuraaga:
After much testing of various servers, I finally found the words for a deployment doc. Happy for any suggestions as this was the least mechanical of any of the docs

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -6,8 +6,8 @@ use a Python application server to run a service.
 ## Application Servers
 
 Connect services are standard ASGI or WSGI applications that can be run with an off-the-shelf Python
-application server. This means generally, any experience in running a Python application can be applied
-to running Connect as-is.
+application server. Generally, any experience in running a Python application can be applied to running
+Connect as-is.
 
 ### HTTP/1.1
 
@@ -41,18 +41,28 @@ When in doubt, if you do not use bidirectional streaming, we recommend one of th
 Connect services are standard ASGI and WSGI applications so any CORS middleware can be used to
 enable it.
 
-For example, with an ASGI application using Starlette:
+For example, with an ASGI application using the [asgi-cors](https://pypi.org/project/asgi-cors/)
+middleware:
 
 ```python
-from starlette.middleware.cors import CORSMiddleware
-from starlette.applications import Starlette
+from asgi_cors import asgi_cors
 
-app = Starlette()
-app.add_middleware(
-    CORSMiddleware,
-    allow_origins=["*"],
-    allow_methods=["GET", "POST"],
-    allow_headers=["*"],
+from greet.v1.greet_connect import GreetServiceASGIApplication
+from server import Greeter
+
+app = GreetServiceASGIApplication(Greeter())
+
+# app is a standard ASGI application - any middleware works as-is
+app = asgi_cors(
+    app,
+    hosts=["https://acme.com"],
+    # POST is used for all APIs except for idempotent unary RPCs that may support GET
+    methods=["GET", "POST"],
+    headers=[
+        "content-type",
+        "connect-protocol-version",
+        "connect-timeout-ms",
+        "x-user-agent",
+    ],
 )
-# Mount your Connect application
 ```


docs/deployment.md
```diff
@@ -0,0 +1,58 @@
...
+from starlette.middleware.cors import CORSMiddleware
+from starlette.applications import Starlette
+
+app = Starlette()
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
+    allow_methods=["GET", "POST"],
+    allow_headers=["*"],
+)
```
stefanvanburen: the `*` seems a little loose to me; could we:

1. for `allow_origins`, add a comment to "replace with your domain" (similar to [cors-go example](https://github.com/connectrpc/cors-go?tab=readme-ov-file#example))
2. add comments to `GET` and `POST` for why they're needed (GET for connect, POST for connect/gRPC)
3. pull in the [headers w/ comments from cors-go](https://github.com/connectrpc/cors-go/blob/fd7420a9f3fd83e76aa6688daf92492210c13fc2/cors.go#L32-L48)?

We may eventually want to have a `cors-go` equivalent in python for ease of adding these headers, but adding these seems like enough for now.

docs/deployment.md
```diff
@@ -0,0 +1,58 @@
...
+application server. This means generally, any experience in running a Python application can be applied
+to running Connect as-is.
+
+### HTTP/1.1
+
+Connect only requires HTTP/2 for bidirectional streaming RPCs - the vast majority of services can be run
+using HTTP/1.1 without issue. Two servers have long been in use with HTTP/1.1 and have a proven track
+record. If you are unsure what server to use, it is generally a safe bet to use one of them:
+
+- [gunicorn](http://www.gunicorn.org/) for WSGI applications
```
i2y, line 18: maybe  `https://` is better?

docs/deployment.md
```diff
@@ -0,0 +1,58 @@
...
+- [uvicorn](https://uvicorn.dev/) for ASGI applications
+
+### HTTP/2
+
```
i2y, line 22: I'm not sure and forgot about daphne.  Should we mention daphne as well?

docs/deployment.md
```diff
@@ -0,0 +1,58 @@
...
+If your service uses bidirectional or otherwise want to use HTTP/2, the above servers will not work.
+HTTP/2 support in the Python ecosystem is still relatively young - servers known to support HTTP/2
+with bidirectional streaming are:
+
+- [granian](https://github.com/emmett-framework/granian)
+- [hypercorn](https://hypercorn.readthedocs.io/en/latest/)
+- [pyvoy](https://github.com/curioswitch/pyvoy)
```
i2y, line 29: I'd love to improve https://github.com/i2y/tsuno so it can be on this list someday! (just a comment)

docs/deployment.md
anuraaga: Unfortunately Daphne's http2 support doesn't work with bidi so I think it's confusing to add to this list. And without it, I think uvicorn is a better choice so left it out.

anuraaga: Unfortunately Daphne's http2 support doesn't work with bidi so I think it's confusing to add to this list. And without it, I think uvicorn is a better choice so left it out.

i2y: Thanks, I see.

docs/deployment.md
```diff
@@ -0,0 +1,58 @@
+# Deployment
+
+After building a Connect service, you still need to deploy it to production. This guide covers how to
+use a Python application server to run a service.
+
+## Application Servers
+
+Connect services are standard ASGI or WSGI applications that can be run with an off-the-shelf Python
+application server. This means generally, any experience in running a Python application can be applied
```
spenczar:
```suggestion
application server. Generally, any experience in running a Python application can be applied
```

docs/deployment.md
```diff
@@ -0,0 +1,58 @@
...
+
+app = Starlette()
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
+    allow_methods=["GET", "POST"],
+    allow_headers=["*"],
+)
+# Mount your Connect application
+```
```
spenczar: I think we should explain how to mount onto Starlette, here.

docs/deployment.md
```diff
@@ -0,0 +1,58 @@
...
+having hung requests or stream ordering issues. pyvoy was built with connect-python in mind but is
+very new and needs more time with real-world applications to verify stability.
+
+Keep the above in mind when picking an HTTP/2 server and let us know how it goes if you give any a try.
+When in doubt, if you do not use bidirectional streaming, we recommend one of the HTTP/1.1 servers.
+
+## CORS
+
+Connect services are standard ASGI and WSGI applications so any CORS middleware can be used to
+enable it.
```
spenczar, line 42: We should probably provide a CORS interceptor, so that we don't give the false impression that users need to use Starlette for CORS support.

docs/deployment.md
anuraaga: I guess they'll be in trouble when Chrome requires it - they don't seem to support https...

docs/deployment.md
anuraaga: I think it's ok to add an interceptor but also feel it's an HTTP concern rather than RPC and could be best left to other libraries. I changed to using asgi-cors to stress this is about using off-the-shelf middleware, not starlette's routing. What do you think?

docs/deployment.md
i2y: When I commented, I thought it connected via https, but it seems I was mistaken. sorry...

i2y (approved):

docs/deployment.md
anuraaga: Going to go ahead and merge this to get it in but happy to follow up on any more suggestions!

PR #58, opened 2025-12-03
Automatically add future annotations import

anuraaga:
From what I understand it's fairly common convention to always use future annotations to have type safety with minimizing runtime overhead. It's the default from Python 3.14 - actually I came up with this when I realized now it's relatively easy writing code that works on 3.14 but crashes on others due to the differing behavior (almost always caught by unit tests but just a developer convenience too).


stefanvanburen:
I'm not sure I know enough about those annotations, but it seems like the current plan is that they'll eventually be deprecated and emit warnings?: https://peps.python.org/pep-0749/#the-future-of-from-future-import-annotations. Would the plan be to remove them when we stop supporting e.g. Python 3.13?

anuraaga:
Yeah that would be the time. That doc seems to acknowledge the common use of these as it seems there wouldn't be a deprecation until 3.13 is EoL, or even longer past that based on the community. That's many many years away and more lenient than most deprecations plans.

> However, many libraries are already using from __future__ import annotations as an elegant way to enable unrestricted forward references in their annotations. If we deprecate the future import immediately, it would be impossible for these libraries to use unrestricted forward references on all supported Python versions while avoiding deprecation warnings

justfile
```diff
@@ -7,7 +7,7 @@ BUF_VERSION := "v1.57.0"
 
 # Format Python files
 format:
-    uv run ruff check --fix .
+    uv run ruff check --fix --unsafe-fixes --exit-zero .
```
anuraaga, line 10: I forgot to comment on this. I have been using unsafe-fixes for a long time and find it great, zero times has it done a wrong fix. That's probably because I work with well behaved code without weird side effects, but I think it applies here too. In the off chance it does do something wrong, manually fixing it would be relatively easy.

spenczar (approved):

Issue #49, opened 2025-11-13
gRPC support timeline


KhanMechAI:
Hi, super keen to get using this, but we'd love to use the gRPC protocol. Do you have a timeline for when support will be added?

payaaam:
It's the latter. I have grpc python server that I want to migrate to connect. I like connect because it allowed me to use the gRPC protocol for service to service calls but the JSOn based connections for client based calls.

We want to write a python server that another backend service connects to. Client would be connect go

PR #68, opened 2025-12-17
Implement server side gRPC protocol

anuraaga:
#49

This implements the server side of gRPC protocol, running tests with pyvoy since it supports trailers. ASGI trailers are an extension defined in the spec and other servers may eventually support them, WSGI there's no definition but pyvoy provides an extension to provide parity between the two.

I know there was some concern about added complexity for supporting gRPC - let me know how it looks. I tried to keep things simpler by noting that gRPC protocol is almost the same as connect's streaming protocol except for error handling - so I didn't touch the unary codepath at all and left it connect-only. If it seems ok, I think it can be great to at least provide server-side support for gRPC to make migration easier than using a proxy server.


src/connectrpc/_server_async.py
```diff
@@ -172,9 +166,14 @@ async def __call__(
             http_method = scope["method"]
             headers = _process_headers(scope.get("headers", ()))
 
-            ctx = _server_shared.create_request_context(
-                endpoint.method, http_method, headers
-            )
+            protocol = negotiate_server_protocol(headers.get("content-type", ""))
```
stefanvanburen: nit: assign `headers.get("content-type", "")` and use it in `msg`?

stefanvanburen (approved):

src/connectrpc/_protocol_grpc.py
```diff
@@ -0,0 +1,153 @@
...
+        case "S":
+            return 1 * 1000
+        case "m":
+            return 1
+        case "u":
+            return 1 / 1000
+        case "n":
+            return 1 / 1000 / 1000
+        case _:
+            msg = "protocol error: timeout has invalid unit '{unit}'"
```
i2y: missing f prefix
`{unit}` won't be interpolated and will appear literally in the error message.

src/connectrpc/_protocol_connect.py
```diff
@@ -0,0 +1,138 @@
...
+CONNECT_PROTOCOL_VERSION = "1"
+CONNECT_HEADER_TIMEOUT = "connect-timeout-ms"
+CONNECT_UNARY_CONTENT_TYPE_PREFIX = "application/"
+CONNECT_STREAMING_CONTENT_TYPE_PREFIX = "application/connect+"
+CONNECT_STREAMING_HEADER_COMPRESSION = "connect-content-encoding"
+CONNECT_STREAMING_HEADER_ACCEPT_COMPRESSION = "connect-accept-encoding"
+CONNECT_UNARY_HEADER_COMPRESSION = "content-encoding"
+CONNECT_UNARY_HEADER_ACCEPT_COMPRESSION = "accept-encoding"
+CONNECT_STREAMING_HEADER_COMPRESSION = "connect-content-encoding"
+CONNECT_STREAMING_HEADER_ACCEPT_COMPRESSION = "connect-accept-encoding"
```
i2y, line 31: some constants are defined twice

  CONNECT_STREAMING_HEADER_COMPRESSION = "connect-content-encoding"
  CONNECT_STREAMING_HEADER_ACCEPT_COMPRESSION = "connect-accept-encoding"
  CONNECT_STREAMING_HEADER_COMPRESSION = "connect-content-encoding"
  CONNECT_STREAMING_HEADER_ACCEPT_COMPRESSION = "connect-accept-encoding"

src/connectrpc/_protocol_grpc.py
anuraaga: Thanks for catching these!

PR #71, opened 2025-12-24
Set default accept-encoding to compression algorithms supported by the client

achille-roussel:
I observed content encoding negotiation issues that I believe was caused by the connect-python client having a hardcoded list of supported encodings despite not necessarily having the corresponding modules loaded. This PR should address it by generating the default Accept-Encoding header based on what the client supports, as detected dynamically at run time.

### DIFF
--- a/src/connectrpc/_compression.py
+++ b/src/connectrpc/_compression.py
@@ -93,7 +93,7 @@ _compressions["identity"] = _identity
 
 # Preferred compression names for Accept-Encoding header, in order of preference.
 # Excludes 'identity' since it's an implicit fallback.
-DEFAULT_ACCEPT_ENCODING_COMPRESSIONS = ["gzip", "br", "zstd"]
+DEFAULT_ACCEPT_ENCODING_COMPRESSIONS = ("gzip", "br", "zstd")
 
 
 def get_compression(name: str) -> Compression | None:


src/connectrpc/_compression.py
```diff
@@ -101,6 +101,16 @@ def get_available_compressions() -> KeysView:
     return _compressions.keys()
 
 
+def get_accept_encoding_compressions() -> list[str]:
+    """Returns compression names suitable for Accept-Encoding header, in preference order.
+
+    This excludes 'identity' since it's an implicit fallback, and returns
+    only compressions that are actually available (i.e., their dependencies are installed).
+    """
+    preferred_order = ["gzip", "br", "zstd"]
```
anuraaga: Let's move this out of the function, maybe DEFAULT_ACCEPT_ENCODING_COMPRESSIONS

src/connectrpc/_compression.py
```diff
@@ -101,6 +101,16 @@ def get_available_compressions() -> KeysView:
...
+    return [name for name in preferred_order if name in _compressions]
```
anuraaga: What do you think about returning the joined string without the intermediate list? A microoptimization, but the list is only benefiting the unit test currently IIUC

anuraaga (commented):
Thanks! This indeed looks like a bug

anuraaga (approved):

PR #76, opened 2026-01-07
Use custom package for grpc status.proto

### LINKED ISSUE #75: status.proto file makes connect-python incompatible with google packages
Simple repro for the problem

In a fresh venv:

```
dgibson@MacBook-Pro-5 /tmp % uv pip install googleapis-common-protos==1.66.0 connect-python==0.7.0 google-cloud-recaptcha-enterprise==1.25.0
Using Python 3.14.2 environment at: badconnect
Resolved 23 packages in 247ms
Installed 14 packages in 61ms
 + cachetools==6.2.4
 + charset-normalizer==3.4.4
 + google-api-core==2.25.2
 + google-auth==2.45.0
 + google-cloud-recaptcha-enterprise==1.25.0
 + grpcio==1.76.0
 + grpcio-status==1.71.2
 + proto-plus==1.27.0
 + pyasn1==0.6.1
 + pyasn1-modules==0.4.2
 + requests==2.32.5
 + rsa==4.9.1
 + typing-extensions==4.15.0
 + urllib3==2.6.2
(badconnect) dgibson@MacBook-Pro-5 /tmp % python
Python 3.14.2 (main, Dec  9 2025, 19:29:30) [Clang 21.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from connectrpc.server import ConnectASGIApplication
>>> from google.cloud import recaptchaenterprise_v1
Traceback (most recent call last):
...
TypeError: Couldn't build proto file into descriptor pool: duplicate symbol 'google.rpc.Status'
```

The above does not reproduce on version 0.6.0, likely because of the status.proto file added in https://github.com/connectrpc/connect-python/commit/45f7e2bdae66559f6acf528e400d53b2ede15935.

anuraaga:
Fixes #75 

I had copied this file from connect-es

https://github.com/connectrpc/connect-es/blob/main/packages/connect/src/protocol-grpc/proto/status.proto

but forgot about the python-specific issue when packages clash. This renames the package so it doesn't conflict with the official Google one. The package name doesn't affect us at all so I tried to be precise with it. Since modifying the file anyways, I removed some non-Python bits from it.

### DIFF
--- a/proto/status.proto
+++ b/proto/status.proto
@@ -14,72 +14,18 @@
 
 syntax = "proto3";
 
-package connectrpc.grpc.compat;
+// This package is for internal use by Connect, and provides no backward
+// compatibility guarantees whatsoever.
+package grpc.status.v1;
 
 import "google/protobuf/any.proto";
 
-// The `Status` type defines a logical error model that is suitable for different
-// programming environments, including REST APIs and RPC APIs. It is used by
-// [gRPC](https://github.com/grpc). The error model is designed to be:
+// See https://cloud.google.com/apis/design/errors.
 //
-// - Simple to use and understand for most users
-// - Flexible enough to meet unexpected needs
-//
-// # Overview
-//
-// The `Status` message contains three pieces of data: error code, error message,
-// and error details. The error code should be an enum value of
-// [google.rpc.Code][google.rpc.Code], but it may accept additional error codes if needed.  The
-// error message should be a developer-facing English message that helps
-// developers *understand* and *resolve* the error. If a localized user-facing
-// error message is needed, put the localized message in the error details or
-// localize it in the client. The optional error details may contain arbitrary
-// information about the error. There is a predefined set of error detail types
-// in the package `google.rpc` which can be used for common error conditions.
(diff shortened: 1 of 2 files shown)


proto/status.proto
```diff
@@ -14,78 +14,18 @@
 
 syntax = "proto3";
 
-package google.rpc;
+// This package is for internal use by Connect, and provides no backward
+// compatibility guarantees whatsoever.
+package grpc.status.v1;
```
anuraaga, line 19: Actually I remembed that Go has the same behavior as Python, so I went ahead and just copied the file as-is from connect-go

https://github.com/connectrpc/connect-go/blob/59cc6973156cd9164d6bea493b1d106ed894f2df/internal/proto/connectext/grpc/status/v1/status.proto#L26

The renamed package is less connectrpc specific, but I figure there's no reason for us to use a different name from it

stefanvanburen (approved):
glad there's a workaround, I hadn't seen (or had forgotten 😅) the connect-go one

PR #77, opened 2026-01-07
Fix async client timeout and cancellation issues

anuraaga:
- Shield httpx response close from timeout so its closing of response iterators happens without interruption
- Separate tasks for executing httpx requests and retrieve responses through a queue
  - httpx does not seem to reliably cancel I/O, so to be able to return from the client quickly when cancelled, this seems like the only pattern to allow it. In some cases the I/O will still make it through to completion, but at least the client isn't blocked on it
- Improve timing of cancel_after_close to be as close to after close as possible. Before it was run right away
- Don't write TLS certs to files on event loop which would block it. I found it while seeing if it was just a busy event loop causing flakiness before the above findings
- Run client conformance tests in parallel to speed them up and do some stress testing of the client, which is what exposed the above bugs

### DIFF
--- a/proto/status.proto
+++ b/proto/status.proto
@@ -14,78 +14,18 @@
 
 syntax = "proto3";
 
-package google.rpc;
+// This package is for internal use by Connect, and provides no backward
+// compatibility guarantees whatsoever.
+package grpc.status.v1;
 
 import "google/protobuf/any.proto";
 
-option go_package = "google.golang.org/genproto/googleapis/rpc/status;status";
-option java_multiple_files = true;
-option java_outer_classname = "StatusProto";
-option java_package = "com.google.rpc";
-option objc_class_prefix = "RPC";
-
-// The `Status` type defines a logical error model that is suitable for different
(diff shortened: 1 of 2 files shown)


conformance/test/test_client.py
```diff
@@ -82,7 +82,7 @@ def test_client_async() -> None:
             "client",
             *_skipped_tests_async,
             "--known-flaky",
-            "Client Cancellation/**",
+            "**/cancel-after-responses",
```
anuraaga, line 85: This is still flaky, I suspect it is an issue in the test harness, not code. I will defer debugging it as already did a lot of debugging and want to get the fixes in

stefanvanburen (approved):
stamping so we can get this in for #79; looks reasonable to me but not as familiar with the asyncio.Queue approach so up to you if you want to wait for a more thorough review.

> httpx does not seem to reliably cancel I/O

Is there an upstream issue we can track? Or ought we open one (maybe hard to pinpoint)?

Issue #83, opened 2026-01-12
Unship docs from this repo?


stefanvanburen:
Now that we have docs in the upstream connectrpc.com repo (on the site: https://connectrpc.com/docs/python/getting-started), should we unship the docs we have here?

PR #84, opened 2026-01-13
Drop docs/ directory

stefanvanburen:
We now have published docs in connectrpc.com repo that we ought to use as the source of truth for connect-python docs; this just drops the docs/ content and related dependencies from this repo.

Completes #83.

Ref: connectrpc/connectrpc.com#312

### DIFF
--- a/justfile
+++ b/justfile
@@ -22,11 +22,11 @@ typecheck:
 # Run unit tests with no extras
 [working-directory('noextras')]
 test-noextras *args:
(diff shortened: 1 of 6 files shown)


justfile
stefanvanburen, line 1: I think `just-lsp` effectively ran `just --fmt --unstable` here; happy to back out the unrelated changes, but figured it was fine.

anuraaga (approved):
Thanks - can you also remove the mkdocs.yaml?

PR #88, opened 2026-01-20
Migrate to pyqwest

anuraaga:
As discussed, this is the spike to replace httpx with pyqwest. Seems like a lot to like here, but of course I'm biased :) Happy to hear what others think.

### DIFF
--- a/conformance/test/client.py
+++ b/conformance/test/client.py
@@ -5,14 +5,11 @@ import asyncio
 import contextlib
 import multiprocessing
 import queue
-import ssl
 import sys
 import time
 import traceback
-from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING, Literal, TypeVar, get_args
 
-import httpx
 from _util import create_standard_streams
 from gen.connectrpc.conformance.v1.client_compat_pb2 import (
     ClientCompatRequest,
@@ -40,9 +37,8 @@ from gen.connectrpc.conformance.v1.service_pb2 import (
     UnimplementedRequest,
 )
 from google.protobuf.message import Message
-from pyqwest import HTTPTransport, SyncHTTPTransport
+from pyqwest import Client, HTTPTransport, SyncClient, SyncHTTPTransport
 from pyqwest import HTTPVersion as PyQwestHTTPVersion
-from pyqwest.httpx import AsyncPyqwestTransport, PyqwestTransport
 
 from connectrpc.client import ResponseMetadata
 from connectrpc.code import Code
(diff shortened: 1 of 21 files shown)


conformance/test/client.py
```diff
@@ -202,32 +163,29 @@ async def client_sync(
...
-                httpx.AsyncClient(transport=transport)
-            )
+    if args:
+        transport = HTTPTransport(**args)
+        # Type parameter for enter_async_context requires coroutine even though
+        # implementation doesn't. We can directly push aexit to work around it.
+        cleanup.push_async_exit(transport.__aexit__)
+        http_client = Client(transport)
+    else:
+        http_client = None
```
anuraaga, line 182: Added coverage of a client without an explicit transport which could have been good before too

src/connectrpc/_client_sync.py
```diff
@@ -415,17 +423,6 @@ def _send_request_bidi_stream(
             raise ConnectError(Code.UNAVAILABLE, str(e)) from e
 
 
-# Convert a timeout with connect semantics to a httpx.Timeout. Connect timeouts
-# should apply to an entire operation but this is difficult in synchronous Python code
```
anuraaga, line 419: With reqwest, the timeout applies from start to end of response read so it works without any weirdness

test/test_lifespan.py
```diff
@@ -42,31 +39,15 @@ async def counting_haberdasher():
             final_count = counter["requests"]
 
     app = HaberdasherASGIApplication(counting_haberdasher())
-    # Use uvicorn since it supports lifespan
```
anuraaga, line 45: pyqwest `ASGITransport` supports lifespan

test/test_errors.py
```diff
@@ -315,75 +338,46 @@ async def make_hat(self, request, ctx):
...
+    response = await client.execute(
         method=method, url=f"http://localhost{path}", content=body, headers=headers
     )
 
-    assert response.status_code == response_status
+    assert response.status == response_status
     assert response.headers == response_headers
 
 
-# To exercise timeouts, can't use mock transports
```
anuraaga, line 329: Mock transports support timeouts

src/connectrpc/_client_async.py
```diff
@@ -173,8 +172,6 @@ async def close(self) -> None:
         """Close the HTTP client. After closing, the client cannot be used to make requests."""
         if not self._closed:
             self._closed = True
-            if self._close_client:
```
anuraaga, line 176: There's nothing left to close. I left he methods though in case we need to add in the future

PR #91, opened 2026-01-21
Prepare for v0.8.0

anuraaga:
Will also get #90 in.

Also won't merge until https://github.com/connectrpc/connectrpc.com/pull/318 is ready to merge


pyproject.toml
```diff
@@ -9,7 +9,7 @@ maintainers = [
     { name = "Yasushi Itoh", email = "i2y.may.roku@gmail.com" },
 ]
 requires-python = ">= 3.10"
-dependencies = ["protobuf>=5.28", "pyqwest>=0.3.2"]
+dependencies = ["protobuf>=5.28", "pyqwest>=0.3.3"]
```
anuraaga, line 12: Sorry about this - I had to fix an issue with closing a running generator in bidi streaming which seems to only have hangs on python 3.12, maybe due to https://github.com/python/cpython/issues/118272. It does mean I will need to be diligent to continue fixing any issues that may come up for users.

On the bright side I am running conformance tests on push

https://github.com/curioswitch/pyqwest/pull/90

Hopefully that's ok, the docs change does have me more pumped on getting to a fully featured connect implementation

PR #92, opened 2026-01-22
Update dev dependencies

anuraaga:
Just dev so not tied to the release


stefanvanburen (approved):

anuraaga:
Looks like several dependencies don't have wheels for 3.14t and get rebuilt, making it slooow. Notable offender is as usual grpcio, a good confirmation of why we have connect-python :)

I think it would be nice to still run this on push rather than PR, but the workarounds needed for that in GitHub actions aren't pretty, will defer for now

PR #103, opened 2026-01-27
Make Compression protocol public and default to gzip-only

anuraaga:
This reworks the interfaces related to compression to better align behavior with other connect implementations and provide more flexibility, and type safety

- Makes `Compression` protocol public. Can be used to implement custom compressions
- Accepts `Compression` instead of `str` throughout for type safety
- Default compression is now gzip-only instead of using dependency detection. `Compression` implementations for zstd and brotli are provided in public API to allow registering them, possibly with custom levels, as users need
- Send compression default to gzip instead of identity to match connect-go
- Removes noextras project now that we don't do dependency detection. Yay

### DIFF
--- a/src/connectrpc/_compression.py
+++ b/src/connectrpc/_compression.py
@@ -7,7 +7,7 @@ from connectrpc.compression.gzip import GzipCompression
 from .compression import Compression
 
 if TYPE_CHECKING:
-    from collections.abc import Sequence
+    from collections.abc import Iterable
 
 
 class IdentityCompression(Compression):
@@ -30,7 +30,7 @@ _default_compressions = {"gzip": _gzip, "identity": _identity}
 
 
 def resolve_compressions(
-    compressions: Sequence[Compression] | None,
+    compressions: Iterable[Compression] | None,
 ) -> dict[str, Compression]:
     if compressions is None:
         return _default_compressions


src/connectrpc/compression/gzip.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import annotations
+
+import gzip
+
+from . import Compression
+
+
+class GZipCompression(Compression):
```
stefanvanburen: minorest of nits: I'd prefer `Gzip` to `GZip` here, to match the stdlib.

conformance/test/gen/connectrpc/conformance/v1/service_connect.py
```diff
@@ -89,7 +91,7 @@ def __init__(
         *,
         interceptors: Iterable[Interceptor] = (),
         read_max_bytes: int | None = None,
-        compressions: Iterable[str] | None = None,
+        compressions: Sequence[Compression] | None = None,
```
stefanvanburen: I might have just missed it reading through, but why do we need a `Sequence` instead of an `Iterable` here?

anuraaga: Thanks for calling this out - in a different project I got feedback that `Sequence` could make sense for an ordered list, i.e. `set` doesn't work then. But on a second thought, I know that such a parameter would be `Iterable` in Java so what works there should work here and switched back to it. If a user passes a `set` anyways, they probably have a strong reason for it for better or worse

PR #114, opened 2026-02-02
Ignore test paths in codecov

anuraaga:


stefanvanburen (approved):
from https://app.codecov.io/github/connectrpc/connect-python/tree/main/, looks like `test/` is already handled, but we may want to skip `src/_gen` as well?

anuraaga:
Great catch, thanks!

Issue #18, opened 2025-09-30
Rewrite protoc-gen-connect-python in Python


stefanvanburen:
We don't *have* to do this, but it seems ideal that we'd implement the plugin in Python to ease contributions to the repo, consolidate on tooling and testing, etc. (Admittedly, Go _is_ pretty nice for writing plugins.)

If we do choose to do this, we could take a look at https://github.com/fischor/protogen-python — the "original" connect-python's plugin was written with this library.

PR #119, opened 2026-02-06
Switch generator to use bufbuild/protoplugin

stefanvanburen:
No strong feeling on landing this, but just wanted to see how our plugin would look if we integrated https://github.com/bufbuild/protoplugin, which eliminates some of the boilerplate of handling the plugin.

Vaguely relates to #18, although I think the language the plugin is written in now is less important than ever.

Also takes a pass at cleaning up the go code and tests.

### DIFF
--- a/protoc-gen-connect-python/generator/generator.go
+++ b/protoc-gen-connect-python/generator/generator.go
@@ -23,20 +23,20 @@ func Handle(ctx context.Context, _ protoplugin.PluginEnv, responseWriter protopl
 
 	conf := parseConfig(request.Parameter())
 
-	files, err := request.FileDescriptorsToGenerate()
+	fileDescriptors, err := request.FileDescriptorsToGenerate()
 	if err != nil {
 		return fmt.Errorf("failed to get file descriptors to generate: %w", err)
 	}
 
-	for _, file := range files {
+	for _, fileDescriptor := range fileDescriptors {
 		// We don't generate any code for non-services
-		if file.Services().Len() == 0 {
+		if fileDescriptor.Services().Len() == 0 {
 			continue
 		}
 
-		name, content, err := generateConnectFile(file, conf)
+		name, content, err := generateConnectFile(fileDescriptor, conf)
 		if err != nil {
-			return fmt.Errorf("failed to generate file %q: %w", file.Path(), err)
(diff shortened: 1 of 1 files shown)


protoc-gen-connect-python/generator/generator.go
```diff
@@ -2,65 +2,49 @@ package generator
...
+func Handle(ctx context.Context, _ protoplugin.PluginEnv, responseWriter protoplugin.ResponseWriter, request protoplugin.Request) error {
+	responseWriter.SetFeatureProto3Optional()
+	responseWriter.SetFeatureSupportsEditions(
+		descriptorpb.Edition_EDITION_PROTO3,
+		descriptorpb.Edition_EDITION_2023,
+	)
 
-	resp.SupportedFeatures = proto.Uint64(uint64(pluginpb.CodeGeneratorResponse_FEATURE_PROTO3_OPTIONAL) | uint64(pluginpb.CodeGeneratorResponse_FEATURE_SUPPORTS_EDITIONS))
-	resp.MinimumEdition = proto.Int32(int32(descriptorpb.Edition_EDITION_PROTO3))
-	resp.MaximumEdition = proto.Int32(int32(descriptorpb.Edition_EDITION_2023))
```
stefanvanburen, line 23: regardless of whether or not we land this, it's probably only a matter of time before we're asked to support [edition 2024](https://protobuf.dev/reference/protobuf/edition-2024-spec/). (similar: grpc/grpc#40639)

stefanvanburen:
I know we've already solved our go -> wheel setup with goreleaser, but looks like https://github.com/simonw/go-to-wheel is a new approach (just wanted to mention!).

protoc-gen-connect-python/generator/generator.go
anuraaga: Cool - I don't mind adding that in this PR, or can be a different one, there doesn't seem to be any new syntax for services as usual

anuraaga (approved):

protoc-gen-connect-python/generator/generator.go
stefanvanburen: I'll add it separately just so it appears separately in the PR list for people looking around.

PR #120, opened 2026-02-06
Drop setup-python from CI

stefanvanburen:
Experimenting with if we need these.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -19,10 +19,6 @@ jobs:
 
       - uses: astral-sh/setup-uv@803947b9bd8e9f986429fa0c5a41c367cd732b41 # v7.2.1
 
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
-        with:
-          python-version-file: "pyproject.toml"
-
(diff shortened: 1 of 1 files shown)


anuraaga (commented):
Thanks @stefanvanburen - I think they're technically not needed but then python becomes whatever happens to be on the github runner at the time, which can be different as the runner image gets updated even within the same label.

For a simple source-only wheel, it may be a bit pedantic but I tend to like having explicit `setup-` actions for all dependencies to still ensure tracking of versions used.

As an aside, I just noticed that GitHub's attestations don't display anything about the runner

https://github.com/connectrpc/connect-python/attestations/17478928

Not sure if it's a display or recording issue, but it seems the workflow is easier to verify for these too.

stefanvanburen:
> Thanks @stefanvanburen - I think they're technically not needed but then python becomes whatever happens to be on the github runner at the time, which can be different as the runner image gets updated even within the same label.
> 
> For a simple source-only wheel, it may be a bit pedantic but I tend to like having explicit `setup-` actions for all dependencies to still ensure tracking of versions used.

Doesn't the python version just end up getting determined by `uv` though? (I guess it ought to just match since I'd expect it to effectively use the same resolution of using the pyproject.toml.)

I hadn't thought about the attestations bit; I don't mind leaving these as-is if they end up being useful in that output down the road.

PR #125, opened 2026-02-09
Mark plugin as supporting edition 2024

stefanvanburen:
Ref: https://github.com/connectrpc/connect-python/pull/119#discussion_r2776064390

### DIFF
--- a/protoc-gen-connect-python/generator/generator_test.go
+++ b/protoc-gen-connect-python/generator/generator_test.go
@@ -95,6 +95,7 @@ func TestGenerateConnectFile(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 			fd, err := protodesc.NewFile(tt.input, nil)
 			if err != nil {
 				t.Fatalf("Failed to create FileDescriptorProto: %v", err)
@@ -200,6 +201,7 @@ func TestGenerate(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 			resp := generate(t, tt.req)
 			if tt.wantErr {
 				if resp.GetError() == "" {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -383,10 +383,10 @@ When exceeded, returns `RESOURCE_EXHAUSTED` error.
 
 ### Proto Editions Support
 
-connect-python supports Proto Editions 2023:
+`protoc-gen-connect-python` supports up to [Protobuf Editions](https://protobuf.dev/editions/overview/) 2024:
```
stefanvanburen, line 386: seemed more right to mention the plugin here, although I know this goes against the consistency of the rest of the README. 🤷

protoc-gen-connect-python/generator/generator.go
```diff
@@ -18,7 +18,7 @@ func Handle(ctx context.Context, _ protoplugin.PluginEnv, responseWriter protopl
 	responseWriter.SetFeatureProto3Optional()
 	responseWriter.SetFeatureSupportsEditions(
 		descriptorpb.Edition_EDITION_PROTO3,
-		descriptorpb.Edition_EDITION_2023,
+		descriptorpb.Edition_EDITION_2024,
```
anuraaga, line 21: Don't we need to keep 2023 here too?

Initially I thought the unit test was somewhat overkill but it alerted me to this so it was pretty good I guess :) Maybe we should `t.Run` across the multiple editions then

stefanvanburen: that tripped me up too, but we're actually declaring the min/max "edition" we support here: https://pkg.go.dev/github.com/bufbuild/protoplugin#ResponseWriter. (I don't love the naming of that method, but I guess it's more obvious once you see the argument names.) So 2023 is implicitly supported.

Even so, I took a shot at making the edition tests table-based and tested both 2023 and 2024, just to have coverage: https://github.com/connectrpc/connect-python/pull/125/commits/91a76ccff78f0f7d7c5b7cde8fa52409c6fe1cd9

anuraaga (approved):

PR #140, opened 2026-02-24
Specify python version to setup-uv

anuraaga:
When reviewing #120 I recall explicitly noticing the setup-python step for the actual tests was still there, making sure they run with the correct python, but I must have misread it. That was important to keep, and I noticed that currently all tests are run with Python 3.12, sorry for missing it.

As a compromise, I've gone ahead and just set the python version to setup-uv without an additional step to see how it goes


stefanvanburen:
My fault in #120, thanks for fixing!

PR #149, opened 2026-03-03
Add exception to metadata interceptors on_end

anuraaga:
Realized while starting OTel middleware - while metadata interceptors aren't meant to support any type of interception, a logging interceptor was definitely in mind when defining them, and that would need access to any raised exceptions.

### DIFF
--- a/test/test_interceptor.py
+++ b/test/test_interceptor.py
@@ -50,12 +50,19 @@ class RequestInterceptor:
 
 
 @pytest.fixture
-def interceptor():
+def client_interceptor():
+    return RequestInterceptor()
+
+
+@pytest.fixture
+def server_interceptor():
     return RequestInterceptor()
 
 
 @pytest_asyncio.fixture
-async def client_async(interceptor: RequestInterceptor):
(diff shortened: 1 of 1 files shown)


test/test_interceptor.py
```diff
@@ -84,6 +106,18 @@ async def test_intercept_unary_async(
...
 
+@pytest.mark.asyncio
+async def test_intercept_unary_async_error(
+    client_async: HaberdasherClient, interceptor: RequestInterceptor
+) -> None:
+    with pytest.raises(ConnectError):
+        await client_async.make_hat(Size(inches=-10))
+    assert (
+        interceptor.result
+        == ["Hello MakeHat and goodbye with error Size must be non-negative"] * 2
```
stefanvanburen: need more coffee, but why are all of these `* 2` again?

anuraaga: I had the same question actually when working on this. Split up the server / client interceptor

stefanvanburen: thanks, now that makes sense!

PR #150, opened 2026-03-03
Add OpenTelemetry middleware

anuraaga:
Initial implement with basic tracing / propagation. Follow-ups will add opt-in metadata attributes and metrics.

Some questions

- Whether to use a separate package at all. As an "OTel guy" I'm happy to go with native instrumentation built into the library as I did with pyqwest (single dep on opentelemetry-api) but started with middleware since it's more common.
- PyPi package name - I went with connectrpc-otel hoping we will publish to connectrpc eventually (not sure the latest on that)
- Whether to use a workspace package vs separate repo. It probably won't really depend on much of conenctrpc and could be better to version separately, but it becomes easier to lose track of the repo

Happy with any direction for these.

https://opentelemetry.io/docs/specs/semconv/rpc/connect-rpc/

### DIFF
--- a/connectrpc-otel/connectrpc_otel/_interceptor.py
+++ b/connectrpc-otel/connectrpc_otel/_interceptor.py
@@ -105,8 +105,6 @@ class OpenTelemetryInterceptor:
                 carrier = ctx.request_headers()
                 parent_otel_ctx = self._propagator.extract(carrier)
 
-        span_kind = SpanKind.CLIENT if self._client else SpanKind.SERVER
-
         rpc_method = f"{ctx.method().service_name}/{ctx.method().name}"
 
         attrs: MutableMapping[str, AttributeValue] = {


connectrpc-otel/connectrpc_otel/_interceptor.py
```diff
@@ -0,0 +1,247 @@
...
+        with self._start_span(ctx) as span:
+            try:
+                return await call_next(request, ctx)
+            except Exception as e:
+                error = e
+                raise
+            finally:
+                self._finish_span(span, error)
+
+    async def intercept_client_stream(
```
anuraaga: I only added unit tests for unary for now since I'll probably avoid the stream-type specific interceptors after https://github.com/connectrpc/connect-python/pull/149

anuraaga:
> I don't know exactly how all the recent python tooling works with this sort of set up. Just wanted to get your thoughts before taking a closer look.

Yeah that should be quite possible - I don't think there's really any special tooling involved, `uv build` only builds the current directory's wheel, and we also point the publish github actions to a folder. We should be able to detect the tag name to control this. The only question is if the Releases page becomes confusing with different project cycles there - for example someone subscribed to repo releases would see connectrpc-otel releases. I would suspect those aren't frequent so it may be fine. 

To clarify the other repos

- > I think we avoided it with Go because it's annoying to do multi-module versioning in a single repo
 
  IIRC, the current proposal `connectrpc-otel/0.1.0` is actually exactly how to do multi-module for Go. So if it was problematic there, maybe it's something to think about.

- > and probably borrowed that approach for the -es repos (

  I think umbrella projects in Node are super-common, and probably wouldn't even be asking about the approach there ;)

connectrpc-otel/connectrpc_otel/_interceptor.py
```diff
@@ -0,0 +1,247 @@
...
+
+    from opentelemetry.util.types import AttributeValue
+
+    from connectrpc.request import RequestContext
+
+REQ = TypeVar("REQ")
+RES = TypeVar("RES")
+
+# Workaround bad typing
+_DEFAULT_TEXTMAP_SETTER = cast("Setter[MutableMapping[str, str]]", default_setter)
```
stefanvanburen, line 44: any upstream issue we ought to follow for this? (I'm not super familiar with the python otel repos.)

connectrpc-otel/connectrpc_otel/_semconv.py
```diff
@@ -0,0 +1,22 @@
+# Vendored in OpenTelemetry semantic conventions for connect-python to avoid
+# unstable imports. We don't copy docstrings since for us they are implementation
+# details and should be obvious enough.
```
stefanvanburen, line 3: Does otel ever stabilize this stuff (/ have a plan to stabilize this stuff)? Just wondering if we need to leave this interceptor at `0.x` until it is (or else consider [always populating old values](https://github.com/connectrpc/otelconnect-go/pull/184#discussion_r1797236086)?)

connectrpc-otel/pyproject.toml
```diff
@@ -0,0 +1,61 @@
+[project]
+name = "connectrpc-otel"
+version = "0.8.1"
```
stefanvanburen: should we land this as `0.1.0`?

connectrpc-otel/pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+description = "OpenTelemetry middleware for connectrpc"
```
stefanvanburen: s/middleware/interceptor?

connectrpc-otel/connectrpc_otel/_interceptor.py
stefanvanburen: > I'll probably avoid the stream-type specific interceptors after #149

hm, what does #149 have to do with that?

connectrpc-otel/connectrpc_otel/_semconv.py
```diff
@@ -0,0 +1,22 @@
...
+ERROR_TYPE: Final = "error.type"
+RPC_METHOD: Final = "rpc.method"
+RPC_RESPONSE_STATUS_CODE: Final = "rpc.response.status_code"
+RPC_SYSTEM_NAME: Final = "rpc.system.name"
+SERVER_ADDRESS: Final = "server.address"
+SERVER_PORT: Final = "server.port"
+
+
+class RpcSystemNameValues(Enum):
+    CONNECTRPC = "connectrpc"
```
stefanvanburen, line 22: since we support gRPC, would we plan on eventually supporting gRPC via this interceptor? or do that elsewhere?

connectrpc-otel/connectrpc_otel/_interceptor.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+class OpenTelemetryInterceptor:
+    """Interceptor to generate telemetry for RPC server and client requests."""
+
+    def __init__(
+        self,
+        *,
+        propagator: TextMapPropagator | None = None,
+        tracer_provider: TracerProvider | None = None,
```
stefanvanburen, line 54: would we also support stuff like a MeterProvider and the various knobs from otelconnect-go? (there's [a lot of Options](https://pkg.go.dev/connectrpc.com/otelconnect#Option) and I'm honestly not clear which ones are something we'd intend to support here, versus may be vestigial.)

src/connectrpc/_server_sync.py
```diff
@@ -213,7 +213,13 @@ def __call__(
                 raise HTTPException(HTTPStatus.NOT_FOUND, [])
 
             http_method = environ["REQUEST_METHOD"]
+            http_scheme = environ.get("wsgi.url_scheme", "http")
             headers = _process_headers(_normalize_wsgi_headers(environ))
+            if ra := environ.get("REMOTE_ADDR"):
+                port = environ.get("REMOTE_PORT", "0")
```
stefanvanburen, line 219: this is WSGI stuff, right?

connectrpc-otel/connectrpc_otel/_interceptor.py
anuraaga: Opening up otel-python repos in an IDE, it's all red ;) I think chasing down typing issues there will be a rabbit hole

connectrpc-otel/connectrpc_otel/_interceptor.py
anuraaga: Yup, will add it with metrics later. Will also consider the options later - some like trust remote, attribute filters, are less common in Python than Go in my experience.

connectrpc-otel/connectrpc_otel/_semconv.py
anuraaga: After many many years yeah :) HTTP and DB are finally stable, RPC is at RC. So I think it's safe to use even in a 1.x if we beat them to it

We're definitely not going to use the old attributes for this new library to keep ourselves sane

connectrpc-otel/connectrpc_otel/_semconv.py
anuraaga: Yup will add it to this interceptor later

connectrpc-otel/pyproject.toml
anuraaga: Went with instrumentation which is the standard term for this in OTel

src/connectrpc/_server_sync.py
anuraaga: Actually it's CGI - WSGI was an extension to CGI, so WSGI servers also commonly populate this (except pyvoy since I didn't notice this until writing this PR, will fix it ;) ). Unfortunately WSGI itself doesn't define this at all.

If a server doesn't populate, we don't get it but that's the best we can do - client IP is pretty important in traces to identify bad actors so it's worth the best effort

connectrpc-otel/connectrpc_otel/_interceptor.py
anuraaga: A lot :) Rewrote to metadata interceptor so the logic is constant across stream types now

connectrpc-otel/connectrpc_otel/_semconv.py
anuraaga: BTW in terms of import, I wish python had two artifacts, one with stable and one with incubating, but they only have one with `_` imports for incubating, and a ton of cruft from before they had the `_` concept. I don't expect their current artifact to ever be 1.x. 

I filed https://github.com/open-telemetry/opentelemetry-python/issues/4952 to see if this may change

PR #151, opened 2026-03-03
Run conformance tests with HTTP/3

anuraaga:
Nothing special connectrpc-wise for supporting HTTP/3 but figured it's worth keeping tabs on our feature set

### DIFF
--- a/conformance/test/client.py
+++ b/conformance/test/client.py
@@ -125,6 +125,8 @@ def pyqwest_client_kwargs(test_request: ClientCompatRequest) -> dict:
             kwargs["http_version"] = PyQwestHTTPVersion.HTTP1
         case HTTPVersion.HTTP_VERSION_2:
             kwargs["http_version"] = PyQwestHTTPVersion.HTTP2
+        case HTTPVersion.HTTP_VERSION_3:
+            kwargs["http_version"] = PyQwestHTTPVersion.HTTP3
     if test_request.server_tls_cert:
         kwargs["tls_ca_cert"] = test_request.server_tls_cert
         if test_request.HasField("client_tls_creds"):


conformance/test/server.py
```diff
@@ -635,40 +639,47 @@ async def serve_pyvoy(
...
-    stdout = proc.stdout
-    assert stdout is not None
-    stdout = _tee_to_stderr(stdout)
+        interface = "asgi"
+        app = "server:asgi_app"
+
+    async def start():
+        # Currently explicit env not accepted but it's safe to set it here.
+        os.environ["READ_MAX_BYTES"] = str(request.message_receive_limit)
+        async with PyvoyServer(
```
anuraaga, line 657: Technically switching to programmatic is a workaround for an issue that pyvoy only prints a single port even if there are several (for HTTP/3, when setting port=0, the TLS and QUIC port end up separate due to Envoy's behavior).

But I figured it cleans things up anyways, it's still a subprocess, just not subprocess of subprocess.

stefanvanburen: is that something you'd want to fix upstream in pyvoy?

conformance/test/server.py
```diff
@@ -635,40 +639,47 @@ async def serve_pyvoy(
...
-    )
-    stdout = proc.stdout
-    assert stdout is not None
-    stdout = _tee_to_stderr(stdout)
+        interface = "asgi"
+        app = "server:asgi_app"
+
+    async def start():
+        # Currently explicit env not accepted but it's safe to set it here.
+        os.environ["READ_MAX_BYTES"] = str(request.message_receive_limit)
```
stefanvanburen, line 656: Is this just so we pipe through the environment that was already going to the subprocess?

conformance/test/server.py
anuraaga: Yup intend to, though not to revert this back, I think programmatic works well here

conformance/test/server.py
anuraaga: Yeah. While I thought it was a rare use case, I will add extra_env or something to `PyvoyServer` to simplify this one

conformance/test/server.py
stefanvanburen: yep, looks reasonable to me.

PR #161, opened 2026-03-09
Add support for gRPC-web

anuraaga:
I think there's pretty low use case for gRPC-web in Python (migrate an existing Python server serving gRPC-web seems uncommon), but it's also very easy to implement when the other two protocols are supported. So went ahead and did it to reduce possible gaps even more.

### DIFF
--- a/conformance/test/test_client.py
+++ b/conformance/test/test_client.py
@@ -28,11 +28,14 @@ _skipped_tests_sync = [
 # expects HTTP/2. It seems reasonable the conformance runner may have some issues here
 # but we don't seem to see failures in other repos so maybe not.
 _flaky_tests = ["--known-flaky", "**/HTTPVersion:1/**/(grpc server impl)/**"]
+
+_skipped_tests = []
 if sys.platform == "darwin":
     # TODO: Investigate HTTP/3 conformance test failures on macOS more.
     # Currently, it seems to be an issue with the conformance runner itself and we see this log message.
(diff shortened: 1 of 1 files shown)


src/connectrpc/_client_async.py
```diff
@@ -92,7 +93,7 @@ def __init__(
         address: str,
         *,
         proto_json: bool = False,
-        grpc: bool = False,
```
anuraaga, line 95: I had knowingly taken the easy way out on this parameter instead of defining a new type hoping grpc-web isn't needed but alas - I think the breaking change should be small / uncommon enough to go with it

src/connectrpc/_protocol_grpc.py
```diff
@@ -274,14 +327,16 @@ def handle_end_message(
...
         return False
 
+    def get_response_trailers(self, response: Response | SyncResponse) -> HTTPHeaders:
+        return response.trailers
+
     def handle_response_complete(
         self, response: Response | SyncResponse, e: ConnectError | None = None
     ) -> None:
-        trailers = response.trailers
-        # Go ahead and feed HTTP trailers regardless of gRPC semantics.
```
anuraaga, line 281: I had decided this not because of any conformance test but it seemed reasonable to allow a user to differentiate between response headers and trailers. gRPC wasn't affected by it but it seems gRPC-web conformance tests verify that `response.headers` are presented to the user as trailers for a trailers-only response

src/connectrpc/_client_async.py
stefanvanburen: seems fine; should we cut a `v0.9.0` in the next few days with this called out? been a little while since the last release.

PR #163, opened 2026-03-10
Add OTel auto-instrumentation

### LINKED ISSUE #164: Bump instruments constraint in OTel auto instrumentation
Currently, the instruments constraint is the current version

https://github.com/connectrpc/connect-python/pull/163/changes#diff-a4524c6583bac282e8d55f718b73d68361e4f173c175a43a625b1b1592af4c25R20

This allows tests to run right now, however, the instrumentation uses APIs that will be released with 0.9.0. It will need to be bumped after #162 and before any connectrpc-otel release. This issue is just to track that.

anuraaga:
While there may be additional knobs to add over time, I think this completes a usable OTel instrumentation package.

Fixes #164

### DIFF
--- a/connectrpc-otel/connectrpc_otel/_instrumentor.py
+++ b/connectrpc-otel/connectrpc_otel/_instrumentor.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
     from connectrpc.interceptor import Interceptor, InterceptorSync
 
-_instruments = ("connect-python>=0.8.0",)
+_instruments = ("connect-python>=0.9.0",)
 
 P = ParamSpec("P")
 R = TypeVar("R")


connectrpc-otel/connectrpc_otel/_instrumentor.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+        register_post_import_hook(self._patch_client, "connectrpc.client")
+        register_post_import_hook(self._patch_server, "connectrpc.server")
+
+    def _uninstrument(self, **kwargs: object) -> None:
+        unwrap("connectrpc.client.ConnectClient", "__init__")
+        unwrap("connectrpc.client.ConnectClientSync", "__init__")
+        unwrap("connectrpc.server.ConnectASGIApplication", "__init__")
+        unwrap("connectrpc.server.ConnectWSGIApplication", "__init__")
```
i2y: I'm not sure, but I think unwrap() with a string argument internally calls import_module(), so _uninstrument() might end up importing connectrpc.client or connectrpc.server as a side effect even if they haven't been loaded yet. Maybe worth guarding with a sys.modules check?

```python
def _uninstrument(self, **kwargs: object) -> None:
    if "connectrpc.client" in sys.modules:
        unwrap("connectrpc.client.ConnectClient", "__init__")
         unwrap("connectrpc.client.ConnectClientSync", "__init__")
    if "connectrpc.server" in sys.modules:
        unwrap("connectrpc.server.ConnectASGIApplication", "__init__")
        unwrap("connectrpc.server.ConnectWSGIApplication", "__init__")
```

connectrpc-otel/pyproject.toml
```diff
@@ -35,10 +35,23 @@ classifiers = [
...
+    "opentelemetry-api>=1.39.1",
+    "opentelemetry-instrumentation==0.60b1",
+]
 
 [dependency-groups]
 dev = [
+    # auto-instrumentation shouldn't include a dependency on the library being instrumented
+    # so it can be packaged in auto-instrumentation distros for general use.
+    #
+    # This library also supports manually initializaing the interceptor.
```
i2y: typo: initializaing → initializing

connectrpc-otel/connectrpc_otel/_instrumentor.py
anuraaga: Thanks! Also sent PR upstream for it

PR #169, opened 2026-03-16
Fix renovate for OTel dependencies

anuraaga:
Sorry my previous attempt was a complete misread of the huge configuration options reference for Renovate. Instead of `matchUpdateTypes`, it's `separateMajorMinor` that allows major/minor to be in a single PR. And then there was also a `ignoreUnstable` flag to bump the `b*` dependencies.

While my intent was for lock file maintenance to be grouped in too, renovate doesn't support that. It was a nice-to-have anyways (main motivation is possibility of lockfile conflicts between PRs, though they're rare)

Less related but the pin in `dependencies`, while intuitive at first since it's a beta dependency, would also cause stuck users as well as stuck renovate, I have loosened it with the comment on why we have to.

And while it seemed to be hitting the subprojects anyways (maybe it resolves the uv workspace implicitly), I went with the renovate docs that suggest a glob for `matchFileNames`.

Checked on my fork (with scheduleMondays removed from it) https://github.com/anuraaga/connect-python/pull/19

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -447,7 +447,7 @@ dev = [
 [package.metadata]
 requires-dist = [
     { name = "opentelemetry-api", specifier = ">=1.39.1" },
-    { name = "opentelemetry-instrumentation", specifier = "==0.60b1" },
+    { name = "opentelemetry-instrumentation", specifier = ">=0.60b1" },
 ]
 
 [package.metadata.requires-dev]


uv.lock
stefanvanburen, line 1: kinda strange to me that this adjusts the uv.lock at the root; is that normal with workspaces or will we eventually have a separate `uv.lock` for `connectrpc-otel`?

connectrpc-otel/pyproject.toml
```diff
@@ -37,7 +37,10 @@ classifiers = [
 ]
 dependencies = [
     "opentelemetry-api>=1.39.1",
-    "opentelemetry-instrumentation==0.60b1",
+    # We do not want to prevent users from updating OTel when connectrpc isn't depending on
+    # the latest. This does introduce a possibility of incompatibilities, but there is no
+    # alternative, and in practice, the APIs we use have been stable for a long time.
+    "opentelemetry-instrumentation>=0.60b1",
```
stefanvanburen, line 43: I was going to say we ought to loosen this to a non "beta" release, but I guess they're [marking all of the `0.x` releases](https://pypi.org/project/opentelemetry-instrumentation/#history) that way 😄

uv.lock
anuraaga: Yeah I think it's normal for workspaces, also languages like Node behave the same. The lock is only for repo development, not package installation, so the idea is to make sure any repo developer has the same experience. While multiple lock files can also achieve that, a single lock file can as well - unless we run into any had conflicts but for dev dependencies that tends to be rare.

connectrpc-otel/pyproject.toml
anuraaga: Welcome to OTel :)

PR #172, opened 2026-03-17
Publish API docs to GitHub pages

anuraaga:
When reviewing #84, I had actually completely forgotten wanting to not drop it at least without knowing how to handle API docs, which I think are helpful to have.

This restores only the API docs back for publishing to GitHub pages. I figure there's no practical way to integrate these into docusaurus so we can add a link to `https://connectrpc.github.io/connect-python/api/` from the official docs. What do you think?


docs/img/logo.svg
stefanvanburen, line 1: this seems to have issues rendering in GitHub, can we just double-check that it renders on the site?

anuraaga: Oops yeah - had given it a try and gave up on it, forgot to delete

PR #175, opened 2026-03-17
Fix server streaming handler not cancelled on client disconnect

### LINKED ISSUE #174: Server streaming handler not  cancelled on client disconnect
**Issue:**

When a client disconnects during a server streaming RPC, the async generator is never closed. The generator continues yielding indefinitely and each yield produces a failed socket write and an asyncio warning:

`WARNING:asyncio:socket.send() raised exception`

**Expected (or at least what I think should be happening):**

the framework should monitor `receive()` for `http.disconnect` and call `aclose()` or something along those lines.

Reproduce:

1. Create server streaming endpoint a generator that yields indefinitely (e.g. a watch rpc)
2. Connect to it with a client
3. Kill the client
4. Observe the server logging the `socket.send() raised exception` indefinitely

Workaround: 

**ASGI middleware that queues `receive()`**

stefanvanburen:
When a client disconnects during a server streaming RPC, the async generator continued yielding indefinitely because receive() was never consulted after the initial request was consumed.

Fix this by spawning a background task in the `EndpointServerStream` case that monitors receive() for http.disconnect. When detected, an Event is set; the streaming loop checks it between yields and raises ConnectError(CANCELED) if set. The response stream is also explicitly `aclose()`'d in the finally block so generator finally-clauses run promptly rather than being deferred to GC.

Fixes #174.

### DIFF
--- a/test/test_roundtrip.py
+++ b/test/test_roundtrip.py
@@ -312,7 +312,6 @@ async def test_server_stream_client_disconnect() -> None:
     # We invoke the ASGI app directly rather than using a real client with a
     # short timeout because a real client could trigger the disconnect before the
     # request body has been fully read, which would be a different code path.
-    # That timing is also particularly hard to control reliably on Windows.
     disconnect_trigger = asyncio.Event()
     response_count = 0
     call_count = 0


src/connectrpc/_server_async.py
```diff
@@ -396,22 +400,48 @@ async def _handle_stream(
...
-                body = writer.write(message)
-                await send(
-                    {"type": "http.response.body", "body": body, "more_body": True}
-                )
+            finally:
+                # Explicitly close the stream so that any generator finally-blocks
+                # run promptly (Python defers async-generator cleanup to GC otherwise).
+                aclose = getattr(response_stream, "aclose", None)
+                if aclose is not None:
+                    await aclose()
```
anuraaga, line 446: IIUC it's possible for this to throw since it's user code and then monitor task would be leaked. When thinking about letting it propagate vs catch and maybe log, I guess a finally in a user generator is still part of the request handler and makes sense to allow to affect the response. How about just reordering then if that makes sense?

src/connectrpc/_server_async.py
```diff
@@ -396,22 +400,48 @@ async def _handle_stream(
...
+                if monitor_task is not None:
+                    monitor_task.cancel()
+                    with contextlib.suppress(CancelledError, Exception):
```
anuraaga: Since we know monitor task never raises, maybe can keep it simply canceled error

test/test_roundtrip.py
```diff
@@ -280,3 +284,73 @@ async def request_stream():
...
+            finally:
+                generator_closed.set()
+
+    app = HaberdasherASGIApplication(InfiniteHaberdasher())
+
+    # Encode a Connect protocol (application/connect+proto) request for Size(inches=10).
+    request_bytes = Size(inches=10).SerializeToString()
+    request_body = struct.pack(">BI", 0, len(request_bytes)) + request_bytes
+
+    disconnect_trigger = asyncio.Event()
```
anuraaga, line 315: I think a client with a timeout of some tens of Ms should be non-flaky, but maybe the windows runner would try extra hard to prove otherwise.

Manually invoking the app seems fine too but let's add a comment then about potential flakiness if using a client with a timeout (we need it to happen after the request has been fully read)?

test/test_roundtrip.py
```diff
@@ -280,3 +284,77 @@ async def request_stream():
...
+    app = HaberdasherASGIApplication(InfiniteHaberdasher())
+
+    # Encode a Connect protocol (application/connect+proto) request for Size(inches=10).
+    request_bytes = Size(inches=10).SerializeToString()
+    request_body = struct.pack(">BI", 0, len(request_bytes)) + request_bytes
+
+    # We invoke the ASGI app directly rather than using a real client with a
+    # short timeout because a real client could trigger the disconnect before the
+    # request body has been fully read, which would be a different code path.
+    # That timing is also particularly hard to control reliably on Windows.
```
anuraaga: Windows was a half joke about the GitHub runners being very slow, not so much about the OS itself ;) Don't need to include this line

stefanvanburen: https://github.com/connectrpc/connect-python/pull/175/commits/8a60c6505375b93b3f0eac3a058a436f6e931062

PR #180, opened 2026-03-23
Separate OTel group in renovate

anuraaga:
I had misunderstood `ignoreUnstable` to be bumping beta to next beta, but it bumps stable to beta as well, which we don't want for most dependencies.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["helpers:pinGitHubActionDigests", "schedule:earlyMondays"],
+  // TODO: Restore "schedule:earlyMondays"
+  "extends": ["helpers:pinGitHubActionDigests"],
   "prHourlyLimit": 0,
   "lockFileMaintenance": {
     "enabled": true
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,6 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": ["helpers:pinGitHubActionDigests", "schedule:earlyMondays"],
+  // TODO: Restore "schedule:earlyMondays"
+  "extends": ["helpers:pinGitHubActionDigests"],
```
stefanvanburen, line 4: I think you can manually trigger a run from the dashboard, if you want to avoid this dance of back and forth :)

anuraaga: Were you able to get that to work? As far as I could tell manual runs also ended up respecting this setting unfortunately, maybe I missed an override though.

stefanvanburen: don't remember exactly, but https://github.com/bufbuild/protovalidate-python/commit/328d768c7197b490d5e2263c0f55a53847169ad6 landed and then I think I ran a manual run two get the subsequent two PRs / commits that day: https://github.com/bufbuild/protovalidate-python/commits/main/. Fine either way.

PR #182, opened 2026-03-23
Add release workflow for connectrpc-otel

anuraaga:


.github/workflows/ci.yaml
```diff
@@ -118,6 +118,9 @@ jobs:
 
       - run: uv build
 
+      - run: uv build
+        working-directory: connectrpc-otel
```
stefanvanburen, line 123: should there be a step below for pushing this to test.pypi?

.github/workflows/release-otel.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+      - uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+
+      - run: uv sync --frozen
+
+      # Builds into root dist directory
+      - run: uv build
+        working-directory: connectrpc-otel
+
+      - name: publish connect-python
```
stefanvanburen:
```suggestion
      - name: publish connectrpc-otel
```

.github/workflows/release-otel.yaml
```diff
@@ -0,0 +1,31 @@
...
+  workflow_dispatch:
+
+permissions:
+  id-token: write
+  attestations: write
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-24.04
```
stefanvanburen, line 15: not to hobnob on this, but any reason not to just use `ubuntu-latest` for all of these?

anuraaga: Only that since Renovate will bump it so I err towards deterministic where possible. I used `-latest` previously with dependabot (somewhat ironic)

.github/workflows/ci.yaml
anuraaga: Copied in the comment from release.yaml since it should have been here it seems :)

stefanvanburen: makes sense, thanks!

PR #183, opened 2026-03-23
Inherit workspace ruff config for connectrpc-otel

anuraaga:
Always forget it

### DIFF
--- a/connectrpc-otel/connectrpc_otel/_instrumentor.py
+++ b/connectrpc-otel/connectrpc_otel/_instrumentor.py
@@ -12,10 +12,9 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Collection
     from types import ModuleType
 
+    from connectrpc.interceptor import Interceptor, InterceptorSync
     from opentelemetry.metrics import MeterProvider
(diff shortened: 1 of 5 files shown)


connectrpc-otel/connectrpc_otel/_instrumentor.py
```diff
@@ -12,11 +12,10 @@
     from collections.abc import Callable, Collection
     from types import ModuleType
 
+    from connectrpc.interceptor import Interceptor, InterceptorSync
     from opentelemetry.metrics import MeterProvider
     from opentelemetry.trace import TracerProvider
 
-    from connectrpc.interceptor import Interceptor, InterceptorSync
```
anuraaga, line 18: @stefanvanburen I think you may have set up imports like this as first-party elsewhere. Personally, I find the default of treating workspace members as third-party ok since I don't see particularly better readability otherwise, and then we get it with no manual config

stefanvanburen: yeah, I really don't mind either way, all a matter of taste/consistency. works for me.

PR #205, opened 2026-04-07
Migrate from just to poe

anuraaga:
Normally I should get consensus in an issue but didn't ;-)

I like poe a lot and use it in all my projects, especially for it's good Windows compatibility. Notably, we had to minimize just usage in CI since we have Wnidows there. The reason we have justfile here is because the previous connect-python repo was using it and I wanted to keep the diff against it minimal, but otherwise would recommend poe.


poe_tasks.toml
```diff
@@ -0,0 +1,116 @@
...
+sequence = [
+  { cmd = "ruff check --fix --unsafe-fixes --exit-zero" },
+  { cmd = "ruff format" }
+]
+
+[tasks.format-toml]
+help = "Apply auto-formatting to TOML files"
+cmd = "tombi format"
+
+[tasks.generate]
```
anuraaga, line 56: Looking forward to the tombi schema supporting the new groups feature (hint hint :) )

https://poethepoet.natn.io/guides/help_guide.html#grouping-tasks

poe_tasks.toml
```diff
@@ -0,0 +1,116 @@
+#:schema https://json.schemastore.org/partial-poe.json
```
anuraaga, line 1: Don't mind putting it directly in pyproject.toml but I tend to split it out to prevent an ever-growing project config file

poe_tasks.toml
```diff
@@ -0,0 +1,116 @@
...
+cmd = "go run github.com/bufbuild/buf/cmd/buf@${BUF_VERSION} generate"
+
+[tasks.generate-test]
+help = "Generate test files"
+cmd = "go run github.com/bufbuild/buf/cmd/buf@${BUF_VERSION} generate"
+cwd = "test"
+
+[tasks.lint]
+help = "Apply all possible linting to files"
+sequence = ["lint-python", "lint-toml", "lint-types"]
```
anuraaga, line 90: I'm always torn on whether lint should run typechecks since it is pretty slow, but I went with it this time since I feel it is the simplest model

.github/workflows/ci.yaml
```diff
@@ -67,37 +67,32 @@ jobs:
...
-        working-directory: conformance
+        run: uv run poe test-conformance ${{ matrix.coverage == 'cov' && '--cov=connectrpc --cov-report=xml' || '' }}
 
       - name: run OTel tests
-        run: uv run pytest ${{ matrix.coverage == 'cov' && '--cov=connectrpc --cov-report=xml' || '' }}
-        working-directory: connectrpc-otel
+        run: uv run poe test-otel ${{ matrix.coverage == 'cov' && '--cov=connectrpc_otel --cov-report=xml' || '' }}
 
       - name: run Go tests
         run: go test ./...
```
anuraaga, line 88: There are a few more that look worth getting into tasks but didn't go too far this time

protoc-gen-connect-python/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/connectrpc/connect-python/protoc-gen-connect-python
 
-go 1.25.0
+go 1.25.7
```
anuraaga, line 3: This is to match the latest buf, or otherwise `go-version-file` selects a too-old Go. While I like to keep this at `0`, it is less important for non-libraries, especially ones commonly used with buf

poe_tasks.toml
stefanvanburen: I'd still love to investigate `ty` someday, maybe when they're a little more stable...

poe_tasks.toml
```diff
@@ -0,0 +1,118 @@
...
+  { cmd = "uv version --active --bump=${_version}" },
+  { cmd = "uv version --active --directory protoc-gen-connect-python --bump=${_version}" }
+]
+args = [
+  { name = "_version", help = "The type of version bump to apply", choices = [
+    "patch",
+    "minor",
+    "major"
+  ] }
+]
```
stefanvanburen, line 18: was just kicking the tires and noticed this doesn't working without an arg supplied; make it a `required` arg?: https://poethepoet.natn.io/guides/args_guide.html

```suggestion
args = [
  { name = "_version", required = true, help = "The type of version bump to apply", choices = [
    "patch",
    "minor",
    "major"
  ] }
]
```

stefanvanburen (approved):
lgtm, one suggestion on the `bump` task

poe_tasks.toml
anuraaga: Don't you mean, when there's a protobuf library that works out of the box with it? ;)

poe_tasks.toml
stefanvanburen: https://github.com/SchemaStore/schemastore/pull/5541

PR #214, opened 2026-04-13
Allow generating only async or sync code

### LINKED ISSUE #209: [Feature Request] Option to disable WSGI/Sync code generation
I will never use the WSGIApplication or SyncClient, so that is just useless code to me. It would be nice to have an option to choose ASGI/async or WSGI/sync and only generate the code you want.

anuraaga:
Fixes #209

Thanks for the idea @jeffsawatzky - I also know codebases that are relatively new and can be async-only, etc, so being able to simplify the generated code looks like a good idea.

I didn't provide separate options for client vs server - I think the use case is codebases, so either you're mixed and should generate all of it or you're not.

The option name itself was challenging - we use `mode=async` `mode=sync` in our unit tests but I don't think `mode` is a good word for a public API. So I went with a tri-state bool on just the option name `async`.

This doesn't touch the import statements for now - in the medium future we will be replacing the plugin framework with something smarter than string replacement, which will also handle import optimization. Until then, I'd like to avoid adding spaghetti to the import statements - it should work great to have ruff automatically optimize the imports after generation.

### DIFF
--- a/protoc-gen-connect-python/generator/config.go
+++ b/protoc-gen-connect-python/generator/config.go
@@ -34,6 +34,11 @@ type Config struct {
 
 	// Imports is how to import dependencies in the generated code.
 	Imports Imports
+
+	// Async indicates whether to only generate asynchronous code. If false,
+	// only synchronous code will be generated. If nil, both synchronous and
+	// asynchronous code will be generated.
+	Async *bool
 }
 
 func parseConfig(p string) Config {
@@ -63,6 +68,15 @@ func parseConfig(p string) Config {
 				cfg.Imports = ImportsAbsolute
 			case "relative":
 				cfg.Imports = ImportsRelative
+			}
+		case "async":
+			switch value {
+			case "true":
+				trueVal := true
+				cfg.Async = &trueVal
+			case "false":
+				falseVal := false
+				cfg.Async = &falseVal
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


protoc-gen-connect-python/generator/config.go
```diff
@@ -64,6 +69,15 @@ func parseConfig(p string) Config {
...
 			}
+		case "async":
+			switch value {
+			case "true":
+				trueVal := true
+				cfg.Async = &trueVal
+			case "false":
+				falseVal := false
+				cfg.Async = &falseVal
+			}
```
stefanvanburen, line 80: first time looking closely at the option parsing, but: should we be failing here on unrecognized option keys/values? (see https://github.com/connectrpc/connect-go/blob/e299aa60ffcf91633e48fcb3dd594cbbb187ec7f/cmd/protoc-gen-connect-go/main.go#L779)

stefanvanburen (approved):

protoc-gen-connect-python/generator/config.go
anuraaga: Yeah probably should - I think we can leave this for now given the future work on codegen

jeffsawatzky:
Thanks for this! @stefanvanburen can we get a release with this?

stefanvanburen:
hey @jeffsawatzky thanks for the poke - we had one in progress in #230, but just realized we ought to [probably bump to a new minor instead](https://github.com/connectrpc/connect-python/pull/230#issuecomment-4336155130) with the breaking change.

PR #218, opened 2026-04-16
Add gunicorn ASGI to conformance tests

stefanvanburen:
Looks like ~~http2~~/ASGI has been added to gunicorn (http2 is still in beta), so figured adding these to our conformance suite was reasonable.

_edit_: Ended up dropping `http2` here as still ran into some issues; maybe still valuable to land the ASGI stuff, though?

Ref: https://gunicorn.org/guides/http2
Ref: https://gunicorn.org/asgi

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,7 +239,7 @@ For ASGI servers:
 
 For WSGI servers:
 
-- [Gunicorn](https://gunicorn.org/) - Python WSGI HTTP Server
+- [Gunicorn](https://gunicorn.org/) - Python WSGI and ASGI HTTP Server
 - [uWSGI](https://uwsgi-docs.readthedocs.io/) - Full-featured application server
 - Any WSGI-compliant server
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -236,10 +236,11 @@ For ASGI servers:
 - [Uvicorn](https://www.uvicorn.org/) - Lightning-fast ASGI server
 - [Daphne](https://github.com/django/daphne) - Django Channels' ASGI server with HTTP/2 support
 - [Hypercorn](https://gitlab.com/pgjones/hypercorn) - ASGI server with HTTP/2 and HTTP/3 support
+- [Gunicorn](https://gunicorn.org/) - Python WSGI and ASGI HTTP Server
 
 For WSGI servers:
 
-- [Gunicorn](https://gunicorn.org/) - Python WSGI HTTP Server
+- [Gunicorn](https://gunicorn.org/) - Python WSGI and ASGI HTTP Server
```
stefanvanburen, line 243: feels a little weird to have this in both places now? 🤷

stefanvanburen:
Still seeing flakes: https://github.com/connectrpc/connect-python/actions/runs/24569274702/job/71837731826#step:8:193, maybe not worth it...

anuraaga (commented):
TBH I have been considering many times to remove daphne from the pytest parameters since it is pretty flaky, and I don't think it's commonly used. We shouldn't add another similar one.

Maybe just keep the changes to server.py and not the rest?

stefanvanburen:
> Maybe just keep the changes to server.py and not the rest?

eh, doesn't feel super valuable to me to have the code in place but not hooked up in the tests. maybe we just keep this in the back of our mind for if the situation ever improves?

Going to close, but will reopen if you feel strongly the other way!

PR #219, opened 2026-04-17
Surface non-Connect handler exceptions to user

### LINKED ISSUE #217: Try to reraise exception from server handlers or otherwise allow logging them
We need to handle exceptions to be able to return a error response to the client, but most likely we should see if we can reraise it so the underlying ASGI (and maybe WSGI) servers can present it to the user, or otherwise allow logging it.

Will need to verify this works even though we finish the response. API-wise, ASGI at least allows us to finish the response and then raise the exception, and we would need to see how servers treat that. For WSGI, since the response is closed by finishing the iterable, it may not be possible and we'd need to consider manual logging instead.

https://github.com/connectrpc/connect-python/discussions/216

anuraaga:
Fixes #217

Ended up folding WSGI in too. Could confirm servers render the exceptions
```
❯ uv run pyvoy example.eliza_service:app
pyvoy listening on 127.0.0.1:8000
Exception in ASGI application
Traceback (most recent call last):
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/applications.py", line 107, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/middleware/errors.py", line 186, in __call__
    raise exc
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/middleware/errors.py", line 164, in __call__
    await self.app(scope, receive, _send)
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/middleware/exceptions.py", line 63, in __call__
    await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/_exception_handler.py", line 53, in wrapped_app
    raise exc
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/_exception_handler.py", line 42, in wrapped_app
    await app(scope, receive, sender)
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/routing.py", line 716, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/routing.py", line 736, in app
    await route.handle(scope, receive, send)
  File "/Users/anuraag/git/connect-python/.venv/lib/python3.14/site-packages/starlette/routing.py", line 462, in handle
    await self.app(scope, receive, send)
  File "/Users/anuraag/git/connect-python/src/connectrpc/_server_async.py", line 224, in __call__
    return await self._handle_unary_connect(
  File "/Users/anuraag/git/connect-python/src/connectrpc/_server_async.py", line 264, in _handle_unary_connect
    response_data = await endpoint.function(request, ctx)
  File "/Users/anuraag/git/connect-python/example/example/eliza_service.py", line 36, in say
    raise TypeError("FOO")
TypeError: FOO
^CShutting down pyvoy...

❯ uv run gunicorn  example.eliza_service_sync:app
[2026-04-17 12:19:27 +0900] [23456] [INFO] Starting gunicorn 25.3.0
[2026-04-17 12:19:27 +0900] [23456] [INFO] Listening at: http://127.0.0.1:8000 (23456)
[2026-04-17 12:19:27 +0900] [23456] [INFO] Using worker: sync
[2026-04-17 12:19:27 +0900] [23457] [INFO] Booting worker with pid: 23457
[2026-04-17 12:19:27 +0900] [23456] [INFO] Control socket listening at /Users/anuraag/.gunicorn/gunicorn.ctl
Exception in WSGI application
Traceback (most recent call last):
  File "/Users/anuraag/git/connect-python/src/connectrpc/_server_sync.py", line 249, in __call__
    return self._handle_unary(
           ~~~~~~~~~~~~~~~~~~^
        environ, start_response, http_method, endpoint, ctx, headers
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/anuraag/git/connect-python/src/connectrpc/_server_sync.py", line 277, in _handle_unary
    response = endpoint.function(request, ctx)
  File "/Users/anuraag/git/connect-python/example/example/eliza_service_sync.py", line 35, in say
    raise TypeError("Foo")
TypeError: Foo
```


src/connectrpc/_server_async.py
```diff
@@ -486,6 +489,8 @@ async def _watch_for_disconnect() -> None:
                         "more_trailers": False,
                     }
                 )
+            if error and not isinstance(error, ConnectError):
```
anuraaga, line 492: We only raise HTTPExceptions before negotiating a response so don't worry here

src/connectrpc/_server_async.py
```diff
@@ -232,7 +232,10 @@ async def __call__(
                     ctx,
                 )
         except Exception as e:
-            return await self._handle_error(e, ctx, send)
+            await self._handle_error(e, ctx, send)
+            if not isinstance(e, (ConnectError, HTTPException)):
+                raise
```
valentimarco, line 237: This should be `raise e` , isn't it?

anuraaga: When inside an `except` block, `raise` is actually enough to reraise the active exception. This is exercised in the `test_async_unhandled_exception_reraised` test

src/connectrpc/_server_sync.py
```diff
@@ -668,3 +671,12 @@ def _drain_request_body(environ: WSGIEnvironment) -> None:
         # server that doesn't do so, so we go ahead and do it ourselves.
         for _ in _read_body(environ):
             pass
+
+
+def _maybe_log_exception(environ: WSGIEnvironment, exc: Exception) -> None:
+    if isinstance(exc, (ConnectError, HTTPException)):
+        return
+    errors: ErrorStream = environ["wsgi.errors"]
+    errors.write(
```
stefanvanburen, line 680: nit: not sure, but should we `errors.flush()` after this?: https://peps.python.org/pep-3333/#input-and-error-streams. Still understanding WSGI's idioms 😅. I'm assuming we ought to as we're the "portable application"...?

test/test_errors.py
```diff
@@ -424,3 +430,152 @@ async def make_hat(self, request, ctx) -> NoReturn:
...
+    assert isinstance(transport.app_exception, TypeError)
+    assert str(transport.app_exception) == "Something went wrong"
+
+
+@pytest.mark.asyncio
+async def test_async_unhandled_exception_reraised_stream() -> None:
+    class RaisingHaberdasher(Haberdasher):
+        def make_similar_hats(
+            self, request: Size, ctx: RequestContext
+        ) -> AsyncIterator[Hat]:
```
stefanvanburen: nit: should this be `NoReturn` typed? (also not `async def`)

test/test_errors.py
```diff
@@ -424,3 +430,152 @@ async def make_hat(self, request, ctx) -> NoReturn:
...
+
+    assert transport.app_exception is None
+
+
+@pytest.mark.asyncio
+async def test_async_connect_exception_not_reraised_stream() -> None:
+    class RaisingHaberdasher(Haberdasher):
+        def make_similar_hats(
+            self, request: Size, ctx: RequestContext
+        ) -> AsyncIterator[Hat]:
```
stefanvanburen: same nit as above

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
   "Topic :: Software Development :: Libraries :: Python Modules",
   "Typing :: Typed",
 ]
-dependencies = ["protobuf>=5.28", "pyqwest>=0.4.1"]
+dependencies = ["protobuf>=5.28", "pyqwest>=0.5.1"]
```
stefanvanburen, line 31: assuming we need this for https://github.com/curioswitch/pyqwest/releases/tag/v0.5.1:

> * WSGI testing transport registers and exposes wsgi.errors to allow testing handlers that use it

Assuming the transparent wheel change in https://github.com/curioswitch/pyqwest/releases/tag/v0.5.0 doesn't need to be called out on our end?

stefanvanburen (approved):
looks good, just some nits / questions

pyproject.toml
anuraaga: I don't think so - we never really called out the native aspect clearly before I think, at least in compatibility terms. The perf aspect is probably too much detail on the consumer side (it's pretty small)

src/connectrpc/_server_sync.py
anuraaga: Good diligence. I'm leaning towards leaving it out though since I can't imagine a server not writing error messages directly, but also the following `For example, to minimize intermingling of data from multiple processes writing to the same error log.` doesn't even make sense to me. Preventing intermingling requires passing a chunk of data in a single `write` call, `flush` doesn't seem to matter. So I interpret it as a PEP bug, which happens some times

stefanvanburen (approved):
lgtm after force-push w/ DCO

PR #230, opened 2026-04-28
Bump version to 0.9.1

anuraaga:
We'll probably have some breaking changes coming up so good to get some fixes out as a patch

PR #232, opened 2026-04-28
Prepare for v0.10.0

stefanvanburen:
Based on [this comment][1], I think our next release ought to be `v0.10.0`.

[1]: https://github.com/connectrpc/connect-python/pull/230#issuecomment-4336155130

### DIFF
--- a/protoc-gen-connect-python/pyproject.toml
+++ b/protoc-gen-connect-python/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "protoc-gen-connectrpc"
-version = "0.9.1"
+version = "0.10.0"
 description = "Code generator for connect-python"
 readme = "README.md"
 requires-python = ">= 3.10"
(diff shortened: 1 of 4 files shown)


anuraaga:
Thanks for checking on this. I was thinking patch for the `async` feature too since it felt relatively minor while not affecting existing code. Once we are out of beta with stability guarantees, this wouldn't be an issue but I was hoping for this release in particular to be a braindead one not requiring reading the release notes (eventually all minor releases should be such).

How does that sound?

stefanvanburen:
@anuraaga no strong feeling, but I wasn't concerned about the async feature, more the breaking change for the customizable codec in https://github.com/connectrpc/connect-python/pull/192? (I don't believe that's made it into a release yet.)

anuraaga (approved):
Thanks for catching that! I read the previous release notes but should have looked through the commits closer, I mixed up compression and codec. Definitely not a patch release.

PR #235, opened 2026-04-29
Ignore content type parameters when matching it

### LINKED ISSUE #234: Server returns 415 for Content-Types with parameters (e.g., application/json; charset=utf-8)
# Summary

The server-side codec lookup in `_protocol_connect.codec_name_from_content_type` strips only the `application/` (or `application/connect+`) prefix and uses the rest verbatim as the codec name. When a client sends a Content-Type with parameters — e.g. `application/json; charset=utf-8` — the lookup key becomes `json; charset=utf-8`, no codec matches, and [_server_async.py:213](https://github.com/connectrpc/connect-python/blob/4859bac2aceb3f3ed80a2c26d67ed7cd5818d37e/src/connectrpc/_server_async.py#L213) (and the _server_sync.py counterparts at [:306](https://github.com/connectrpc/connect-python/blob/4859bac2aceb3f3ed80a2c26d67ed7cd5818d37e/src/connectrpc/_server_sync.py#L306)/[:437](https://github.com/connectrpc/connect-python/blob/4859bac2aceb3f3ed80a2c26d67ed7cd5818d37e/src/connectrpc/_server_sync.py#L437)) raises `HTTPException(HTTPStatus.UNSUPPORTED_MEDIA_TYPE)`.

Per [RFC 9110 §8.3.2](https://www.rfc-editor.org/rfc/rfc9110.html#name-charset) the charset parameter is part of the standard media-type grammar and the value is case-insensitive so servers are expected to tolerate it.

# Reproduction
```
# Works:
curl -X POST -H 'Content-Type: application/json' -d '{}' \
  https://my-server/<service>/<method>
# → 401/200 — codec resolves

# Fails:
curl -X POST -H 'Content-Type: application/json; charset=utf-8' -d '{}' \
  https://my-server/<service>/<method>
# → HTTP 415 Unsupported Media Type
```

Hits in production when a browser or proxy adds `; charset=utf-8` to the Connect-ES request (we observed this on a Connect-ES --> Connect-Python path even though @connectrpc/connect's request-header.js builds the Content-Type without parameters).

# Expected behavior

Something matching the reference Go implementation, ie. canonicalizeContentType calls mime.ParseMediaType, discards/canonicalizes parameters, and matches on the bare media type.

# Suggested fix

In `_protocol_connect.codec_name_from_content_type`, parse the input via Python's media-type parser before stripping the prefix:

```
from email.message import Message

def codec_name_from_content_type(content_type: str, *, stream: bool) -> str:
    msg = Message()
    msg["content-type"] = content_type
    base = msg.get_content_type()  # e.g. "application/json"
    prefix = (
        CONNECT_STREAMING_CONTENT_TYPE_PREFIX
        if stream
        else CONNECT_UNARY_CONTENT_TYPE_PREFIX
    )
    if base.startswith(prefix):
        return base[len(prefix):]
    return base
```

Thanks for all the work on the project. Happy to send a PR if it's useful. :)

anuraaga:
Fixes #234 

I had intended to have similar handling of content type as Go but realized they treat it quite differently on server vs client side and missed that. This takes the simplest approach of just splitting on `;` and ignoring parameters - I don't think anyone will miss the use case of being able to match on them with custom codecs. Note that custom codecs and header matching isn't defined in the protocol spec and i.e. connect-es don't support custom codecs FWICT. We may not have either but it is important for supporting different protobuf implementations.

For reference, Go, which does canonicalization in a way that technically custom codecs can match on parameters. Digging in, it looks like their somewhat complicated loop is to avoid two passes for both finding a `;` and uppercase (2 passes goes from 9ns to 11ns on my machine). As we can't write such fast loops in Python, I take the typical approach of always calling `.strip().lower()`.

https://github.com/connectrpc/connect-go/blob/a5a6c30f3776b06ae05a66ab3bdd2d60c46db6db/protocol.go#L361

### DIFF
--- a/src/connectrpc/_protocol_connect.py
+++ b/src/connectrpc/_protocol_connect.py
@@ -5,7 +5,7 @@ import struct
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Any, TypeVar
 
-from ._codec import CODEC_NAME_JSON, CODEC_NAME_JSON_CHARSET_UTF8, Codec
+from ._codec import CODEC_NAME_JSON, Codec
 from ._compression import IdentityCompression, negotiate_compression
 from ._envelope import EnvelopeReader, EnvelopeWriter
 from ._protocol import (
@@ -36,6 +36,9 @@ CONNECT_HEADER_PROTOCOL_VERSION = "connect-protocol-version"
 CONNECT_PROTOCOL_VERSION = "1"
 CONNECT_HEADER_TIMEOUT = "connect-timeout-ms"
 CONNECT_UNARY_CONTENT_TYPE_PREFIX = "application/"
+CONNECT_UNARY_CONTENT_TYPE_JSON = (
+    f"{CONNECT_UNARY_CONTENT_TYPE_PREFIX}{CODEC_NAME_JSON}"
+)
 CONNECT_STREAMING_CONTENT_TYPE_PREFIX = "application/connect+"
 CONNECT_UNARY_HEADER_COMPRESSION = "content-encoding"
 CONNECT_UNARY_HEADER_ACCEPT_COMPRESSION = "accept-encoding"
@@ -46,7 +49,16 @@ CONNECT_STREAMING_HEADER_ACCEPT_COMPRESSION = "connect-accept-encoding"
 _DEFAULT_CONNECT_USER_AGENT = f"connectrpc/{__version__}"
 
 
+def _normalize_content_type(content_type: str) -> str:
+    # content-type can have parameters, most commonly charset. Our support codecs,
+    # binary and JSON are always either non-text or utf-8 and the parameters are not
+    # important for matching to a codec. A custom codec could conceivably need to
+    # match on parameters, but we will reconsider that if it is ever asked for.
+    return content_type.partition(";")[0].strip()
+
+
 def codec_name_from_content_type(content_type: str, *, stream: bool) -> str:
+    content_type = _normalize_content_type(content_type)
     prefix = (
         CONNECT_STREAMING_CONTENT_TYPE_PREFIX
         if stream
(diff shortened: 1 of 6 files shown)


src/connectrpc/_server_async.py
```diff
@@ -213,7 +213,7 @@ async def __call__(
                 codec_name = protocol.codec_name_from_content_type(
                     headers.get("content-type", ""), stream=not is_unary
                 )
-            codec = self._codecs.get(codec_name.lower())
```
anuraaga, line 216: I also noticed a couple of spots of random lowercasing and consolidated them, with some cleanup to WSGI

src/connectrpc/_protocol_connect.py
```diff
@@ -226,12 +238,10 @@ def create_request_context(
     def validate_response(
         self, request_codec_name: str, status_code: int, response_content_type: str
     ) -> None:
+        response_content_type = _normalize_content_type(response_content_type)
```
stefanvanburen, line 241: nit that we're double-normalizing the content type between here and `codec_name_from_content_type` below, but no real suggestions on a refactor — probably fine to leave.

test/test_http.py
```diff
@@ -0,0 +1,62 @@
...
+    "application/json; charset=UTF-8",
+    "application/json;charset=utf-8",
+    "application/json;    charset=utf-8",
+    "application/json; charset=utf-8; version=1",
+    "application/JSON",
+]
+
+
+@pytest.mark.parametrize("header", _charset_content_type_cases)
+def test_json_charset_content_type(header: str) -> None:
```
stefanvanburen, line 36: would it also be good to flex streaming in these tests? I see we have `validate_stream_response`; presumably a well-behaved connect server wouldn't have an issue, though...

stefanvanburen (approved):
lgtm, assuming we don't need to worry about the grpc case (e.g. `application/grpc+json; charset=utf-8`)? I don't see that mentioned in https://connectrpc.com/docs/multi-protocol/ / https://chromium.googlesource.com/external/github.com/grpc/grpc/+/HEAD/doc/PROTOCOL-HTTP2.md but perhaps in practice it ... could happen? 😄

test/test_http.py
anuraaga: Almost thought it's too difficult to be worth it but realized I could get by with a normal client and transport middleware to force content-type

anuraaga:
While not too worried about it, for gRPC, I think we are covered anyways by the grammar definition not including paramaters, except for the `{custom}` case which we are explicitly calling out as not suppporting for now.

`Content-Type → “content-type” “application/grpc” [(“+proto” / “+json” / {custom})]`

PR #242, opened 2026-05-06
Switch type-checking from `pyright` to `ty`

stefanvanburen:
Figured I'd see how far along `ty` is; feels pretty good to me. Was able to eliminate more of the type checker ignores along the way.

### DIFF
--- a/conformance/test/test_client.py
+++ b/conformance/test/test_client.py
@@ -6,8 +6,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 import pytest
-
-from ._util import VERSION_CONFORMANCE, coverage_env, maybe_patch_args_with_debug
+from _util import VERSION_CONFORMANCE, coverage_env, maybe_patch_args_with_debug
 
 if TYPE_CHECKING:
     from coverage import Coverage
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,10 +47,12 @@ dev = [
   "gunicorn==25.3.0",
   "hypercorn==0.18.0",
   "poethepoet==0.45.0",
-  "pyright[nodejs]==1.1.409",
   "pyvoy==0.3.0",
   "ruff==0.15.12",
   "tombi==0.10.2",
+  "ty==0.0.34",
+  "types-grpcio==1.0.0.20260408",
+  "types-protobuf==7.34.1.20260503",
   "typing_extensions==4.15.0",
   "uvicorn==0.46.0",
   "zstandard==0.25.0",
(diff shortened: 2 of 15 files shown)


pyproject.toml
```diff
@@ -79,30 +81,19 @@ docs = ["mkdocstrings-python==2.0.3", "zensical==0.0.39"]
...
-
-  # TODO: Work out the import issues to allow it to work.
-  "conformance/**",
 ]
 
+[tool.ty.environment]
+# Conformance test files import generated code as `from gen.connectrpc...` and
+# sibling modules like `_util`/`_cov_embed`. Pytest and the launched server/client
+# subprocesses put `conformance/test/` on sys.path; teach ty to do the same.
+extra-paths = ["conformance/test"]
```
stefanvanburen: this, plus dropping the `__init__.py` and reworking the imports above let us typecheck these.

pyproject.toml
```diff
@@ -47,10 +47,12 @@ dev = [
   "gunicorn==25.3.0",
   "hypercorn==0.18.0",
   "poethepoet==0.45.0",
-  "pyright[nodejs]==1.1.409",
   "pyvoy==0.3.0",
   "ruff==0.15.12",
   "tombi==0.10.2",
+  "ty==0.0.34",
+  "types-grpcio==1.0.0.20260408",
+  "types-protobuf==7.34.1.20260503",
```
anuraaga, line 55: Thanks - didn't realize there was this hatch for getting protobuf working with it

src/connectrpc/_server_async.py
```diff
@@ -220,7 +220,11 @@ async def __call__(
                     [("Accept-Post", "application/json, application/proto")],
                 )
 
-            if is_unary and isinstance(protocol, ConnectServerProtocol):
+            # Re-check via isinstance (rather than reusing `is_unary`) so the type
```
anuraaga: I couldn't reproduce the issue - pushed a change to revert it

conformance/test/test_client.py
```diff
@@ -6,8 +6,7 @@
 from typing import TYPE_CHECKING
 
 import pytest
-
-from ._util import VERSION_CONFORMANCE, coverage_env, maybe_patch_args_with_debug
+from _util import VERSION_CONFORMANCE, coverage_env, maybe_patch_args_with_debug
```
anuraaga, line 9: I guess this doesn't hurt though adding `conformance` to the search paths would also make sense, I guess any pytest root should be added to it in principle.

Reminds me of the days when conformance was a separate pyproject, with renovate working properly we may consider eventually going back to that to reduce pythonpath worries

src/connectrpc/_client_async.py
```diff
@@ -31,9 +30,7 @@
 from .protocol import ProtocolType
 
 try:
-    from asyncio import (
-        timeout as asyncio_timeout,  # pyright: ignore[reportAttributeAccessIssue]
-    )
+    from asyncio import timeout as asyncio_timeout  # ty: ignore[unresolved-import]
```
anuraaga: Using `ImportError` is a very old habit of mine for shims, but ever since `version_info` tuple I guess using that may be more idiomatic (typing-extensions uses it extensively), can consider rewriting to that to be properly type aware

anuraaga (approved):
One random side effect is using the default pyright version in VSCode for normal setups (not using ty language server which will continue to be the norm for a long time). Amazingly one patch version of drift (it loads 1.1.408) seems to cause this pyright bug in widening `bytes` to buffer protocol. Not our problem I would say but figured worth a share.

Image

src/connectrpc/_server_async.py
stefanvanburen: thanks, I thought I double-checked this; oh well!

src/connectrpc/_client_async.py
stefanvanburen: switched in https://github.com/connectrpc/connect-python/pull/242/commits/f3b7d1843d738cb727026e248f5fb05790b18eeb!

anuraaga (approved):

PR #250, opened 2026-05-22
Update unary-get query parameter to match spec ordering

oliversun9:
This is a follow-up to  https://github.com/connectrpc/connectrpc.com/pull/357


anuraaga (approved):
Thanks! I applied a small cleanup to

- Remove docs about Python dict behavior which is pretty well known by now
- Combine assertions on the same test case to a single test

https://github.com/connectrpc/connect-python/pull/250/commits/e742bc4fe96de81dd115ffc8d6a2dba26ec1a814

PR #264, opened 2026-06-08
Generator: emit RequestContext[Input, Output] for service handlers

### LINKED ISSUE #260: generator: emit RequestContext[Input, Output] for service handlers
## Summary

The protoc plugin generates server handler signatures with a **bare** `ctx: RequestContext`.
Because the runtime class is `class RequestContext(Generic[REQ, RES])`, a bare
`RequestContext` is an implicit `RequestContext[Any, Any]`. The generator already knows the
input/output message types per method, so it should emit the precise
`ctx: RequestContext[<Input>, <Output>]`.

## Motivation

**1. Strict mypy is unusable.** Consumers running `disallow_any_generics = true` get an error
on **every** generated handler:

```
error: Missing type parameters for generic type "RequestContext"  [type-arg]
```

**2. The bare annotation silently erases `ctx.method` and its fields.** The runtime classes
are wired together by the same two type vars:

```python
# request.py
class RequestContext(Generic[REQ, RES]):
    @property
    def method(self) -> MethodInfo[REQ, RES]: ...   # passes REQ/RES through

# method.py
class MethodInfo(Generic[REQ, RES]):
    input: type[REQ]
    output: type[RES]
```

So `REQ`/`RES` bound on `ctx` propagate to `ctx.method` and on to `ctx.method.input`/`.output`.
With the bare annotation they bind to "unknown/Any", and that erasure cascades (revealed types
below are from `ty`; mypy reports the `Any` equivalents):

| Expression          | bare `ctx: RequestContext`     | precise `ctx: RequestContext[Size, Hat]` |
| ------------------- | ------------------------------ | ---------------------------------------- |
| `ctx.method`        | `MethodInfo[Unknown, Unknown]` | `MethodInfo[Size, Hat]`                  |
| `ctx.method.input`  | `Unknown`                      | `type[Size]`                             |
| `ctx.method.output` | `Unknown`                      | `type[Hat]`                              |

The information is fully present **at runtime** — the generator constructs
`MethodInfo(input=Size, output=Hat, ...)` with the concrete classes — so the only thing
discarding it for the type checker is the bare annotation on `ctx`. Concretely,
`ctx.method.input()` is `Size` under the precise form but degrades to `Any`/`Unknown` under
the bare form, silently disabling type checking on it.

**3. It contradicts the project's own docs.** The
[getting-started guide](https://connectrpc.com/docs/python/getting-started) already documents
the precise form (`ctx: RequestContext[GreetRequest, GreetResponse]`) — the generator should
match it.

## Current behavior

```python
class Haberdasher(Protocol):
    async def make_hat(
        self, request: haberdasher__pb2.Size, ctx: RequestContext
    ) -> haberdasher__pb2.Hat: ...

    # server-streaming
    def make_similar_hats(
        self, request: haberdasher__pb2.Size, ctx: RequestContext
    ) -> AsyncIterator[haberdasher__pb2.Hat]: ...
```

## Proposed behavior

```python
class Haberdasher(Protocol):
    async def make_hat(
        self, request: haberdasher__pb2.Size,
        ctx: RequestContext[haberdasher__pb2.Size, haberdasher__pb2.Hat],
    ) -> haberdasher__pb2.Hat: ...

    # server-streaming — ctx params are the BARE messages, NOT AsyncIterator-wrapped
    def make_similar_hats(
        self, request: haberdasher__pb2.Size,
        ctx: RequestContext[haberdasher__pb2.Size, haberdasher__pb2.Hat],
    ) -> AsyncIterator[haberdasher__pb2.Hat]: ...
```

For **streaming** methods, the request/return are wrapped in `AsyncIterator`/`Iterator`, but
`ctx`'s type params must remain the bare message types — exactly how the template already
builds `MethodInfo(input=..., output=...)`, which the runtime types as
`RequestContext._method: MethodInfo[REQ, RES]`. So `RequestContext[Input, Output]` keeps `ctx`
in agreement with `MethodInfo` for unary **and** streaming methods.

## Scope

- Affects only the server-side `Protocol` handler signatures (both ASGI/async and WSGI/sync).
  Client stubs don't take `ctx` and are unchanged.
- **Runtime behavior is unchanged** — this is purely a type-annotation precision fix.
- No new imports needed (`RequestContext` and the message types are already referenced by the
  template).

## Compatibility

Backward compatible at runtime. The annotation becomes more specific; existing handlers
annotated with bare `RequestContext` continue to satisfy the Protocol (gradual typing).
Regenerating updates committed fixtures (e.g. `conformance/.../service_connect.py`).

## Validation

- Update the Go template golden tests.
- Regenerate fixtures (`buf generate` / repo's `generate` task).
- Add a typing regression test asserting `ctx` is parameterized with the bare message types
  for unary and streaming methods.

## References

- Runtime: `src/connectrpc/request.py` (`RequestContext(Generic[REQ, RES])`),
  `src/connectrpc/method.py` (`MethodInfo(Generic[REQ, RES])`).
- Template: `protoc-gen-connect-python/generator/template.go`.
- Docs already showing the precise form: <https://connectrpc.com/docs/python/getting-started>

smparkes:
The plugin generated a bare `ctx: RequestContext` for every service handler. Since RequestContext is Generic[REQ, RES], that is an implicit RequestContext[Any, Any]: it trips strict-mypy consumers (disallow_any_generics -> "Missing type parameters for generic type RequestContext") and erases ctx.method.input/.output to Any even though the runtime MethodInfo carries the concrete classes.

Emit the precise `ctx: RequestContext[<Input>, <Output>]` using the raw message types, so the context's params agree with MethodInfo(input=..., output=...) for unary and streaming methods alike. Streaming request/return stay wrapped in AsyncIterator/Iterator, but the context params are the bare messages.

Scope: server-side Protocol handlers only (async + sync); client stubs are unchanged. Runtime behavior is unchanged -- purely a type-annotation precision fix.

Tests (TDD, written failing first):
- Go: TestConnectTemplateRequestContextTypeParams in template_test.go.
- Python: test/test_request_context_typing.py asserts ctx is parameterized with the bare message types for unary and streaming handlers, via get_type_hints + get_origin/get_args.

Regenerated fixtures: test/haberdasher_connect.py, conformance service_connect.py, example eliza_connect.py.

Closes #260


protoc-gen-connect-python/generator/template_test.go
```diff
@@ -90,3 +90,78 @@ func TestConnectTemplate(t *testing.T) {
 		})
 	}
 }
+
+// TestConnectTemplateRequestContextTypeParams verifies that server-side Protocol handlers
+// receive a precisely-parameterized context, ctx: RequestContext[Input, Output], rather than
+// a bare RequestContext (which is an implicit RequestContext[Any, Any] and trips strict
+// mypy's disallow_any_generics). The type params must always be the BARE message types and
+// must agree with MethodInfo(input=..., output=...) -- even for streaming methods, whose
+// request/return are wrapped in AsyncIterator/Iterator but whose context params are not.
```
anuraaga:
```suggestion
```

protoc-gen-connect-python/generator/template_test.go
```diff
@@ -90,3 +90,78 @@ func TestConnectTemplate(t *testing.T) {
...
+	}
+
+	var buf bytes.Buffer
+	if err := ConnectTemplate.Execute(&buf, vars); err != nil {
+		t.Fatalf("Template execution failed: %v", err)
+	}
+	result := buf.String()
+
+	// The precise, parameterized form must be emitted for both the async (ASGI) and the
+	// sync (WSGI) Protocol handlers; each appears once per variant.
```
anuraaga:
```suggestion
```

protoc-gen-connect-python/generator/template_test.go
```diff
@@ -90,3 +90,78 @@ func TestConnectTemplate(t *testing.T) {
...
+		}
+	}
+
+	// The context's type params must be the bare message types, never the stream-wrapped
+	// forms -- they have to agree with MethodInfo(input=..., output=...).
+	for _, bad := range []string{"RequestContext[AsyncIterator", "RequestContext[Iterator"} {
+		if strings.Contains(result, bad) {
+			t.Errorf("context type params must be bare messages, not stream-wrapped: found %q\n--- got ---\n%s", bad, result)
+		}
+	}
```
anuraaga:
```suggestion
```

test/test_request_context_typing.py
```diff
@@ -0,0 +1,60 @@
+"""Regression test for the precision of the generated ``ctx`` annotation.
```
anuraaga: Let's delete this file

anuraaga (commented):
Thanks @smparkes - just need to remove some unnecessary macguffins

test/test_request_context_typing.py
smparkes: Sure. Wasn't sure if you wanted it. I error on the side of more tests for PRs. Plus since we don't have a mypy test here, I wanted to convince myself we were getting the right metadata in there.

anuraaga: The template test is generally enough and can eyeball the changes well to the Python this PR. I'll see if we can make the ty check more strict or such so we don't need this one

protoc-gen-connect-python/generator/template_test.go
smparkes: Double checking (I read this suggestion backwards at first):

Just want to check that you'd rather this not be here. It does double check against an future change that emits both typed and untyped for some reason.

I would keep it (revert my accepting the suggestion) but am fine with your preference.

test/test_request_context_typing.py
smparkes: Ack. (FWIW, I did look to see if ty could handle this; my understanding is that it's not as strict as mypy: no --disallow-any-generics equivalent.)

protoc-gen-connect-python/generator/template_test.go
anuraaga: Yeah we should remove these - if we had a negative assertion for every positive assertion, it becomes too much.

smparkes:
Changes made. LMK if you want the double check tests back or any other issues.
Thanks!

anuraaga (approved):

smparkes:
CI failures are flakes? Not near my code and conf tests run fine locally. Just checking ...

anuraaga:
Thanks!

PR #268, opened 2026-06-15
Manage pytest as normal dependencies

anuraaga:
Not sure but I think we moved these to constraint dependencies to reduce dependabot issues - with renovate, all pytest versions should get bumped together and not have any issue. On the flip side, renovate can't bump constraint-dependencies, so we end up with not getting updates to these.

Also scoped the "don't bump project.dependencies" to the two published Python packages, so that example can be bumped (for starlette).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -243,11 +243,6 @@ exclude = [
 # subprocesses put `conformance/test/` on sys.path; teach ty to do the same.
 extra-paths = ["conformance/test"]
 
-[tool.uv]
-constraint-dependencies = [
-  "coverage==7.13.2",
-]
-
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -248,10 +246,6 @@ extra-paths = ["conformance/test"]
 [tool.uv]
 constraint-dependencies = [
   "coverage==7.13.2",
```
stefanvanburen: > On the flip side, renovate can't bump constraint-dependencies, so we end up with not getting updates to these.

should we go ahead and delete/move this one as well then?

stefanvanburen (approved):

PR #274, opened 2026-06-24
Migrate to protobuf-py and rename repo to connect-py

anuraaga:
This brings the long awaited update to target the new protobuf-py implementation. Along with it, we will rename the repo to connect-py so this incorporates docs updates for that as well.

- Default dependency is protobuf-py
- A `compat` package with codecs targeting `google.protobuf` is included, and automatically used by gencode if passing the option `protobuf=google`
- Rename `protoc-gen-connect-python` to `protoc-gen-connectrpc` which could have happened before too
- Docs are generally updated to follow connect-es's pattern of using just `Connect` in prose, and `connect-py` only in repo links (before we used `Connect-Python` in prose)
- I will backfill unit tests for the codegen plugin (it is mostly tested implicitly except for the `io` option). I owe a testing utility for this in protobuf-py


.gitignore
```diff
@@ -118,13 +118,7 @@ venv.bak/
 
 .vscode
 
-# generated Go binary
-protoc-gen-connect-python/protoc-gen-connect-python
-
-# goreleaser archives
 out/
```
stefanvanburen: probably a couple of others we could delete while we're here?

conformance/test/client.py
```diff
@@ -457,11 +472,11 @@ def send_unary_request_sync(
...
                                 raise ValueError(msg)
-                        if test_request.cancel.after_close_send_ms:
-                            await asyncio.sleep(
-                                test_request.cancel.after_close_send_ms / 1000.0
-                            )
-                            task.cancel()
+                        if test_request.cancel:
+                            match test_request.cancel.cancel_timing:
+                                case Oneof("after_close_send_ms", after_close_send_ms):
+                                    await asyncio.sleep(after_close_send_ms / 1000.0)
```
stefanvanburen, line 479: nit: should this also `await request_closed.wait()` similar to the async case below? (looks like this was already the case in the code, just think it ought to match?)

src/connectrpc/_codec.py
```diff
@@ -39,35 +69,27 @@ def name(self) -> str:
...
+        return message.to_binary()
 
     def decode(self, data: bytes | bytearray, message: V) -> V:
-        # ParseFromString accepts the buffer protocol at runtime, but typeshed
-        # declares only `bytes`. Skipping the conversion avoids a copy on the
-        # streaming decode path (see _envelope.py). Tracked upstream in
-        # https://github.com/python/typeshed/issues/9006.
-        message.ParseFromString(data)  # ty: ignore[invalid-argument-type]
-        return message
+        return message.__class__.from_binary(data)  # TODO: fix type
```
stefanvanburen: guessing this needs an upstream fix?

stefanvanburen (approved):
lgtm, just some nits. 🚢

src/connectrpc/_codec.py
anuraaga: Oops nice catch! Nope it's a fix to the codec type here to allow using protobuf-py's classmethod style better that I had initially punted.

PR #292, opened 2026-07-06
Add gRPC reflection support

### LINKED ISSUE #206: [Feature] Support gRPC Reflection
as title.

## Possible Implementations
Maybe have a method call `enable_reflection()` when creating the  `ASGIApplication` e/o `WSGIApplication`.

anuraaga:
Fixes #206

This adds implementations of gRPC reflection, following the practice of other connect implementations, and even grpc-python actually of accepting the services to return rather than introspecting the running server (which could be a starlette router, etc).

Because protobuf-py allows vendored codegen, I took the approach of just generating the official protos via BSR, and did the same for status.proto too which we previously vendored.

As a quite small file in the end, it feels like overkill to publish a separate pypi package, but it is definitely independent functionality. Open to thoughts.


src/connectrpc/_gen/google/rpc/__init__.py
stefanvanburen, line 1: nit: can we mark the `_gen/` directory here as linguist-generated in the .gitattributes?

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,253 @@
...
+        The services in the passed in descriptors will be made available for reflection.
+        For [DescFile], this means all the services declared in that file. Note that file
+        dependencies are also automatically added and available for resolution by name,
+        but the service list will only reflect the descriptors passed in.
+
+        Args:
+            *descs: The descriptors to make available for reflection.
+
+        Returns:
+            A new instance of [ServerReflectionService], for use with [ServerReflectionASGIApplication].
```
stefanvanburen: This says it's for use with `ServerReflectionASGIApplication` but it's actually used in the 4 subclasses below, maybe could be reworded?

separately, we may need to enable something else in the docs build to linkify these?:

Image

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,253 @@
...
+    seen: set[str] = set()
+    service_names: set[str] = set()
+    for desc in descs:
+        if isinstance(desc, DescFile):
+            fds.extend(fd for fd in _file_descriptor_with_dependencies(desc, seen))
+            service_names.update(d.type_name for d in desc.services)
+        else:
+            fds.extend(fd for fd in _file_descriptor_with_dependencies(desc.file, seen))
+            service_names.add(desc.type_name)
+    return Registry(*fds), list(service_names)
```
stefanvanburen: should we sort the list of service_names for deterministic results?

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,253 @@
...
+        original_req = ServerReflectionRequest.from_binary(req.to_binary())
+    res = ServerReflectionResponse(valid_host=req.host, original_request=original_req)
+    match req.message_request:
+        case Oneof("file_by_filename", filename):
+            desc = registry.file(filename)
+            res.message_response = _create_file_response(req, desc, seen)
+        case Oneof("file_containing_symbol", symbol):
+            desc: DescFile | None = None
+            for d in registry:
+                if isinstance(d, (DescFile, DescExtension)):
```
stefanvanburen: why do we skip extensions?

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,253 @@
...
+) -> (
+    Oneof[Literal["file_descriptor_response"], FileDescriptorResponse]
+    | Oneof[Literal["error_response"], ErrorResponse]
+):
+    if not desc:
+        return Oneof(
+            "error_response",
+            ErrorResponse(
+                error_code=_CODE_NOT_FOUND,
+                error_message=f"symbol {req.message_request} not found",
```
stefanvanburen: this will leak the `Oneof` representation of the underlying `message_request`, should we use `req.message_request.value` instead?

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,253 @@
...
+    seen: set[str],
+    registry: Registry,
+    service_names: list[str],
+) -> ServerReflectionResponse:
+    original_req = req
+    if isinstance(req, ServerReflectionAlphaRequest):
+        # HACK: The v1alpha request type is binary compatible with v1, but will fail to serialize as a submessage
+        # of an unrelated type, so to otherwise allow the rest of the implementation to be shared, we do this
+        # conversion here.
+        original_req = ServerReflectionRequest.from_binary(req.to_binary())
```
stefanvanburen: I can't really say I understand this code - is it still valid? I don't see a test that's hitting this branch but maybe I'm missing it. I think this is perhaps already handled in callers?

src/connectrpc/grpcreflect/_service.py
anuraaga: Ah good catch - I think linking to the app is important so took the user-focused, maintainer-challenging approach of just duplicating the four docstrings.

And while I thought the `[]` form was working fine for us before, I noticed that currently on protobufpy.com it's not working there either. Maybe a zensical change - I guess we'll need to stick to absolute links for robustness. For now did it for this PR and will do a sweep later.

On the bright side, this reminded me to add the protobuf inventory link so we link back to protobufpy.com!

src/connectrpc/grpcreflect/_service.py
anuraaga: Indeed, this was from a hacky `cast` approach which I abandoned since the proper type-safe one wasn't too bad, deleted

src/connectrpc/grpcreflect/_service.py
anuraaga: Not sure why I thought to do that, fixed

src/connectrpc/grpcreflect/_service.py
anuraaga: Just simplified it, the original request is returned anyways

src/connectrpc/_gen/google/rpc/__init__.py
stefanvanburen: still should do this I think 😄

connectrpc-grpcreflect/connectrpc_grpcreflect/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+from __future__ import annotations
+
+__all__ = [
+    "ServerReflectionASGIApplication",
+    "ServerReflectionAlphaASGIApplication",
+    "ServerReflectionAlphaService",
+    "ServerReflectionAlphaServiceSync",
+    "ServerReflectionAlphaWSGIApplication",
+    "ServerReflectionService",
```
stefanvanburen, line 11: any plan to export `ServerReflectionClient` et al for client impls (perhaps as a follow-up)? seems like we ought to provide a client w/ similar version negotiation logic to grpcreflect-go

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,300 @@
...
+        case Oneof("file_by_filename", filename):
+            desc = registry.file(filename)
+            res.message_response = _create_file_response(desc, seen)
+        case Oneof("file_containing_symbol", symbol):
+            desc: DescFile | None = None
+            for d in registry:
+                if isinstance(d, DescFile):
+                    continue
+                if d.type_name == symbol:
+                    desc = d.file
```
stefanvanburen: nit: should break after we've found the type

```suggestion
                if d.type_name == symbol:
                    desc = d.file
                    break
```

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,300 @@
...
+
+def _resolve_registry(
+    descs: tuple[DescFile | DescService, ...],
+) -> tuple[Registry, list[str]]:
+    fds: list[DescFile] = []
+    seen: set[str] = set()
+    service_names: set[str] = set()
+    for desc in descs:
+        if isinstance(desc, DescFile):
+            fds.extend(fd for fd in _file_descriptor_with_dependencies(desc, seen))
```
stefanvanburen:
```suggestion
            fds.extend(_file_descriptor_with_dependencies(desc, seen))
```

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,300 @@
...
+            service_names.update(d.type_name for d in desc.services)
+        else:
+            fds.extend(fd for fd in _file_descriptor_with_dependencies(desc.file, seen))
```
stefanvanburen:
```suggestion
            fds.extend(_file_descriptor_with_dependencies(desc.file, seen))
```

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,300 @@
...
+    # Avoid recursion since not difficult
+    queue: deque[DescFile] = deque()
+    queue.extend(root.dependencies)
+    while queue:
+        file = queue.popleft()
+        if file.name in seen:
+            continue
+        if file.name not in seen:
+            seen.add(file.name)
+            yield file
```
stefanvanburen: nit: we've already `continue`'d above, can dedent here instead
```suggestion
        seen.add(file.name)
        yield file
```

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,300 @@
...
+        for dep in file.dependencies:
+            queue.append(dep)
```
stefanvanburen:
```suggestion
        queue.extend(file.dependencies)
```

src/connectrpc/grpcreflect/_service.py
```diff
@@ -0,0 +1,300 @@
...
+    service_names: list[str],
+) -> ServerReflectionResponse:
+    res = ServerReflectionResponse(valid_host=req.host, original_request=req)
+    match req.message_request:
+        case Oneof("file_by_filename", filename):
+            desc = registry.file(filename)
+            res.message_response = _create_file_response(desc, seen)
+        case Oneof("file_containing_symbol", symbol):
+            desc: DescFile | None = None
+            for d in registry:
```
stefanvanburen: I think this is missing methods, fields, and enumeration values which IIUC should be a part of the grpcreflect "spec". I think we should have test cases for e.g. `connectrpc.example.Hat.name` and `connectrpc.example.Haberdasher.MakeHat`, etc.

Something like the following might work?:

```python
def _find_file_for_symbol(registry: Registry, symbol: str) -> DescFile | None:
    desc = (
        registry.message(symbol)
        or registry.enum(symbol)
        or registry.service(symbol)
        or registry.extension(symbol)
    )
    if desc:
        return desc.file
    # The symbol may instead name a member of a type: a service method
    # (pkg.Service.Method), a message field or oneof (pkg.Message.field), or
    # an enum value, which per proto scoping rules lives in the enum's parent
    # scope (pkg.VALUE for a top-level enum in package pkg).
    parent, _, member = symbol.rpartition(".")
    if not member:
        return None
    if svc := registry.service(parent):
        if any(m.name == member for m in svc.methods):
            return svc.file
        return None
    if (msg := registry.message(parent)) and (
        any(f.name == member for f in msg.fields)
        or any(o.name == member for o in msg.oneofs)
    ):
        return msg.file
    # An enum value of a message-nested enum also has a message as its parent
    # scope, so always finish with the enum value scan.
    for d in registry:
        if (
            isinstance(d, DescEnum)
            and d.type_name.rpartition(".")[0] == parent
            and any(v.name == member for v in d.values)
        ):
            return d.file
    return None
```

stefanvanburen (commented):
getting closer, just found a couple more things.

re:
> it feels like overkill to publish a separate pypi package

I guess it does make me a tad nervous to include it in the base package as breakages could really bite us, and I imagine we'd like to get to connect-py 1.0 somewhat soon after protobuf-py makes it there. doesn't seem like a huge lift to split it out? could be worth considering.

sorry for the late thoughts; appreciate putting up with the rounds of reviews here.

src/connectrpc/_gen/google/rpc/__init__.py
anuraaga: Sorry forgot git add

connectrpc-grpcreflect/connectrpc_grpcreflect/__init__.py
anuraaga: Didn't even think of providing a client since it's always been grpcurl or something. Let's keep this in backlog, I suspect users aren't needing the client

src/connectrpc/grpcreflect/_service.py
anuraaga: Thanks for noticing this - I was relying on the Golang doc for FindDescriptorByName, which are no docs 😅 So I looked at the implementation and also the reflection.proto, it seems to indeed include method names and enum values, but not field names so went with that

test/buf.gen.yaml
```diff
@@ -4,6 +4,9 @@ plugins:
     out: .
   - local: protoc-gen-connectrpc
     out: .
+  # TODO: Consider extracting a testing package to share haberdasher among packages.
+  - local: protoc-gen-py
```
anuraaga, line 8: This seemed simplest for now

pyproject.toml
```diff
@@ -41,6 +41,9 @@ Repository = "https://github.com/connectrpc/connect-py"
 
 [dependency-groups]
 dev = [
+  # Install all packages to be available in docs building, etc.
```
anuraaga, line 44: Found this is probably a good idea

docs/api.md
```diff
@@ -19,3 +19,16 @@
         show_submodules: true
 
 ::: connectrpc.compat
+
+
+::: connectrpc_grpcreflect
```
anuraaga, line 24: Probably a good time to split pages but will do that later

anuraaga (commented):
Thanks - wasn't too worried on stability since the unique API, accepting descriptors, seemed pretty simple. But still, it's probably worth the split, and matches what grpcio does anyways

connectrpc-grpcreflect/connectrpc_grpcreflect/__init__.py
stefanvanburen: sounds good - I suspect eventually we might want to provide something similar to [grpcreflect-go's client](https://github.com/connectrpc/grpcreflect-go/blob/main/client.go), which is a little complex — mostly around the fallback behavior. I'm sure users could just use the generated stuff for one or the other, but might be nice to wrap it ourselves (eventually).

stefanvanburen: sounds good - I suspect eventually we might want to provide something similar to [grpcreflect-go's client](https://github.com/connectrpc/grpcreflect-go/blob/main/client.go), which is a little complex — mostly around the fallback behavior. I'm sure users could just use the generated stuff for one or the other, but might be nice to wrap it ourselves (eventually).

connectrpc-grpcreflect/README.md
```diff
@@ -0,0 +1,36 @@
+# connectrpc-grpcreflect
+
+gRPC reflection services to support tools such as [grpcurl](https://github.com/fullstorydev/grpcurl)
```
stefanvanburen: nit: I know grpcurl is the O.G., but we may want to at least mention `buf curl` or other GUI/terminal clients eventually.

.gitattributes
```diff
@@ -0,0 +1,4 @@
+* text=auto eol=lf
+
+src/connectrpc/_gen linguist-generated=true
+conformance/test/gen linguist-generated=true
```
stefanvanburen, line 4:
```suggestion
conformance/test/gen linguist-generated=true
connectrpc-grpcreflect/connectrpc_grpcreflect/_gen linguist-generated=true
```

stefanvanburen (approved):
lgtm

PR #298, opened 2026-07-11
Add note about http client ownership to client docstrings

dongjiang1989:
## Summary
- `close()` now calls the underlying HTTP client's `close()`/`aclose()` to release connection pool resources (with graceful fallback when the method is not available on the HTTP client)
- All `execute_*` methods now raise `RuntimeError` when called on a closed client
- `_consume_single_response` ensures `stream.aclose()` is called in a `finally` block to prevent async generator resource leaks
- Extracted `_check_closed()` helper to avoid duplication across execute methods

## Test plan
- [x] New unit tests in `test/test_client_close.py` covering all three fixes (13 tests)
- [x] Existing test suite passes (231 passed, 2 pre-existing unrelated failures in `test_example.py`)

### DIFF
--- a/src/connectrpc/_client_async.py
+++ b/src/connectrpc/_client_async.py
@@ -175,7 +175,32 @@ class ConnectClient:
         self._execute_bidi_stream = execute_bidi_stream
 
     async def close(self) -> None:
-        """Close the HTTP client. After closing, the client cannot be used to make requests."""
+        """Mark the client as closed.
+
+        After closing, the client cannot be used to make requests.
+
+        Note:
+            This method does not close the underlying HTTP client. If you provided
+            your own HTTP client, you are responsible for closing it yourself.
+
+        Example::
+
+            import asyncio
+            from pyqwest import Client
+            from my_service import MyServiceClient
+
+            async def main():
+                http_client = Client()
+                client = MyServiceClient("http://localhost", http_client=http_client)
+                try:
+                    # Use the client...
+                    pass
+                finally:
+                    await client.close()  # Marks connect client as closed
+                    await http_client.aclose()  # You must close the HTTP client yourself
+
+            asyncio.run(main())
+        """
         if not self._closed:
             self._closed = True
 
--- a/src/connectrpc/_client_async.py
+++ b/src/connectrpc/_client_async.py
@@ -189,6 +189,7 @@ class ConnectClient:
             from pyqwest import Client
             from my_service import MyServiceClient
 
+
             async def main():
                 http_client = Client()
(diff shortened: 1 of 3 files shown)


src/connectrpc/_client_async.py
```diff
@@ -178,6 +178,9 @@ async def close(self) -> None:
         """Close the HTTP client. After closing, the client cannot be used to make requests."""
         if not self._closed:
             self._closed = True
+            aclose = getattr(self._http_client, "aclose", None)
```
anuraaga: We default to a shared client when not provided, similar to Go clients and it should not be closed. When user provided, we don't own the client and can't close it (it may be shared). So I don't think we should be closing the client here (we used to a while back when not using a shared client by default). We can add more docs to the docstrings about this and examples of closing on the user side probably.

src/connectrpc/_client_async.py
```diff
@@ -440,13 +451,14 @@ async def _yield_single_message(message: REQ) -> AsyncIterator[REQ]:
...
+    try:
+        response = await anext(stream)
+    except StopAsyncIteration:
+        raise ConnectError(Code.UNIMPLEMENTED, "unary response has zero messages") from None
+    try:
+        await anext(stream)
+    except StopAsyncIteration:
+        return response
+    finally:
+        await stream.aclose()
```
anuraaga: The generator is the one returned by _bidi_request which doesn't have a close method, a private implementation detail. Did you notice an issue relating to this?

anuraaga:
Thanks @dongjiang1989 - some questions

src/connectrpc/_client_async.py
dongjiang1989: Thanks @anuraaga 
Just add examples in docs

src/connectrpc/_client_async.py
dongjiang1989: Thanks @anuraaga 
Got it, Remove it down

anuraaga:
Thanks @dongjiang1989 - I have moved the docs to the constructor and used context manager

anuraaga (approved):

dongjiang1989:
Thanks @anuraaga

anuraaga:
@dongjiang1989 Please don't force-push after I've cleaned up / approved. I restored to my state and will merge after CI again

dongjiang1989:
> @dongjiang1989 Please don't force-push after I've cleaned up / approved. I restored to my state and will merge after CI again

Got it.

anuraaga:
Thanks!

PR #299, opened 2026-07-11
Avoid unnecessary copies in IdentityCompression and reuse protocol singletons

dongjiang1989:
## Summary
- **IdentityCompression**: `compress`/`decompress` now return `bytes` input as-is (`data if isinstance(data, bytes) else bytes(data)`) instead of always calling `bytes(data)`, eliminating a redundant allocation on the hot path when no compression is used.
- **Protocol singletons**: `negotiate_server_protocol` returns module-level singleton instances (`_CONNECT_PROTOCOL`, `_GRPC_PROTOCOL`, `_GRPC_WEB_PROTOCOL`) instead of creating new stateless protocol instances on every request. The three server protocol classes (`ConnectServerProtocol`, `GRPCServerProtocol`, `GRPCWebServerProtocol`) are verified to be fully stateless (no `__init__`, no instance variables).
- Added unit tests for both optimizations (20 new test cases).

## Test plan
- [x] `test/test_compression.py::TestIdentityCompression` — 9 cases covering `bytes` identity, `bytearray`/`memoryview` copy semantics, and empty input.
- [x] `test/test_protocol_server.py::TestNegotiateServerProtocol` — 11 cases covering all content-type routing, singleton identity (`is`), correct types, and distinct instances.
- [x] Full test suite: 240 passed.

Signed-off-by: dongjiang1989 <dongjiang1989@126.com>


anuraaga:
Thanks @dongjiang1989 - I have applied some cleanups to the tests, notably deleting the protocol singleton test case since it's a minor optimization that doesn't need a regression test, copying buffers in identity compression on the other hand is a bug

anuraaga (approved):

dongjiang1989:
Thanks @anuraaga

PR #303, opened 2026-07-15
Add dependabot.yml for automated dependency updates

dongjiang1989:
## Summary

- Add `.github/dependabot.yml` to enable automated dependency updates via Dependabot
- Configure weekly updates for GitHub Actions and Python dev dependencies
- Production dependencies (`protobuf-py`, `pyqwest`, `opentelemetry-*`, `connectrpc`, etc.) remain manually managed, consistent with the existing `renovate.json` policy
- Only the root directory is configured for `pip` ecosystem since all uv workspace members share a single `uv.lock`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,52 @@
+version: 2
+updates:
+  # GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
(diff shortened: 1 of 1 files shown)


anuraaga:
Hi @dongjiang1989 - we use renovate instead of dependabot for dependencies for various reasons. While bug fixes and similar are appreciated, for high level changes it is probably worth filing an issue. Please take care in causing too much review churn. Thanks.

PR #305, opened 2026-07-17
Render docstrings and expose service descriptor

anuraaga:
This renders docstrings for service and method very similar to how protobuf-py does for other types. It also adds a `desc()` classmethod for accessing the descriptor also following protobuf-py's model. I debated against a module-level constant since there isn't anything async or sync about the descriptor, but it felt cleanest going with this consistency instead.


protoc-gen-connectrpc/protoc_gen_connectrpc/__init__.py
```diff
@@ -509,6 +534,39 @@ def _message_ident(method: DescMethod, message: DescMessage, options: Options) -
...
+def _generate_desc_method(f: File, service: DescService, options: Options) -> None:
+    if options.protobuf == _ProtobufOption.PY:
+        f.print("@classmethod")
+        with f.scope("def desc(cls) -> ", _DESC_SERVICE, ":"):
+            with f.doc("Returns the descriptor for this service."):
+                pass
+            f.print(
+                "return next(s for s in ",
+                service.file,
+                ".desc().services if s.type_name == '",
```
anuraaga, line 560: I considered threading in the index within the descriptor we see here, but while it should be quite rare, it could be possible for the pb to change especially ordering of items without regenerating connect for some reason. Didn't seem worth avoiding this more flexible matching.

conformance/test/gen/connectrpc/conformance/v1/service_connect.py
stefanvanburen, line 1: I think we're maybe missing the `/**/*` in our own .gitattributes to properly mark these files as generated (think [this approach in protobuf-py](https://github.com/bufbuild/protobuf-py/blob/a8d39098cfb288932df0fbff287ac0bcaa9085ac/.gitattributes#L8-L9) should work).

connectrpc-grpcreflect/connectrpc_grpcreflect/_gen/grpc/reflection/v1alpha/reflection_connect.py
```diff
@@ -26,9 +28,18 @@
 
 
 class ServerReflection(Protocol):
+    """"""
```
stefanvanburen: maybe should skip completely empty docstrings?

protoc-gen-connectrpc/protoc_gen_connectrpc/__init__.py
```diff
@@ -509,6 +534,39 @@ def _message_ident(method: DescMethod, message: DescMessage, options: Options) -
...
+        if len(text) > 0:
+            text += "\n\n"
+        text += comments.trailing.removesuffix("\n")
+    if len(text) > 0:
+        text += "\n"
+
+    for line in text.splitlines():
+        # Comments in protobuf often start with a space, this
+        # leads to weird indentation in the generated docstring, so we remove it.
+        f.print(line.removeprefix(" "))
```
stefanvanburen: seems reasonable to me as a first pass — I think both `buf format` and some of the other plugins have had to make tweaks over time to make this right for the ecosystem. Not too concerned with a little churn in the comments if we need to revisit (and not coming up with a better suggestion now).

stefanvanburen (approved):
looks good, I think we ought to just skip empty docstrings though(?) (and would be nice to fix the .gitattributes as well)

PR #312, opened 2026-07-28
Enforce most lint rules

anuraaga:
This refreshes the lint config, notably lining it up mostly with protobuf-py

### DIFF
--- a/connectrpc-otel/connectrpc_otel/_interceptor.py
+++ b/connectrpc-otel/connectrpc_otel/_interceptor.py
@@ -57,14 +57,17 @@ class OpenTelemetryInterceptor:
         meter_provider: MeterProvider | None = None,
         client: bool = False,
     ) -> None:
-        """Creates a new OpenTelemetry interceptor.
+        """Create a new OpenTelemetry interceptor.
 
         Args:
             propagator: The OpenTelemetry TextMapPropagator to use. If not
                 provided, the global default will be used.
             tracer_provider: The OpenTelemetry TracerProvider to use. If not
                 provided, the global default will be used.
+            meter_provider: The OpenTelemetry MeterProvider to use. If not
+                provided, the global default will be used.
             client: Whether this interceptor is for a client or server.
+
         """
         self._client = client
         tracer_provider = tracer_provider or get_tracer_provider()
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,118 +110,52 @@ skip-magic-trailing-comma = true
 docstring-code-format = true
 
 [tool.ruff.lint]
-extend-select = [
-  # Same order as listed on https://docs.astral.sh/ruff/rules/
-  "YTT",
-  "ANN",
-  "ASYNC",
-  "S",
-  "FBT",
-  "B",
-  "A",
-  "COM818",  # Other comma rules are handled by formatting
-  "C4",
-  "DTZ",
-  "T10",
-  "EM",
-  "EXE",
-  "FA",
-  "ISC",
-  "ICN",
-  "LOG",
-  "G",
-  "INP",
-  "PIE",
-  "T20",
-  "PYI",
-  "PT",
-  "Q004",  # Other quote rules are handled by formatting
-  "RSE",
-  "RET",
-  # This rule is a bit strict since accessing private members within this package can be
-  # useful to reduce public API exposure. But it is important to not access private members
-  # of external dependencies so we enable it and ignore the locations where we need to.
-  "SLF",
-  "SIM",
-  "SLOT",
-  "TID",
-  # TODO: Update TODOs then enable
-  # "TD",
-  "TC",
-  "ARG",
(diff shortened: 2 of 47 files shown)


src/connectrpc/compat/__init__.py
```diff
@@ -1,3 +1,5 @@
+"""Support for the google.protobuf Protobuf implementation with Connect."""
```
stefanvanburen: nit: already forgetting, but would these render better in API docs if we rst'd up the `google.protobuf` as a package somehow? (maybe just wrapping in backticks would be better, from https://mkdocstrings.github.io/griffe/reference/docstrings/#google-style?)

src/connectrpc/compression/gzip.py
```diff
@@ -9,18 +11,22 @@ class GzipCompression(Compression):
...
+        """Create a new GzipCompression.
 
         Args:
             level: Compression level to use.
+
         """
         self._level = level
 
     def name(self) -> str:
+        """Return the compression name for GZip."""
```
stefanvanburen: the lack of symmetry in capitalization between `GZip` and `GzipCompression` is slightly annoying, ah well :^)

src/connectrpc/_protocol.py
```diff
@@ -102,7 +101,7 @@ def from_exception(exc: Exception) -> ConnectWireError:
     def from_response(response: FullResponse) -> ConnectWireError:
         try:
             data = response.json()
-        except Exception:
+        except Exception:  # noqa: BLE001
```
stefanvanburen: nit: are we able to tighten up these `except Exception`s here? I think in perhaps this case we could use `ValueError`, and maybe the following one with something like `with contextlib.suppress(ValueError):`.

Separately, I wonder if we ought to take a pass at switching from `noqa` -> `ruff: ignore` (https://docs.astral.sh/ruff/rules/noqa-comments/).

src/connectrpc/_protocol_connect.py
stefanvanburen, line 1: nit: seems weird that this file was marked as generated; I think the `_connect.py` suffix is tripping off the match. Could rename / tighten up that match somehow?

pyproject.toml
```diff
@@ -110,118 +110,52 @@ skip-magic-trailing-comma = true
...
+  "PLR0914",  # Too many local variables
+  "PLR0915",  # Too many statements
+  "PLR0916",  # Too many boolean expressions
+  "PLR0917",  # Too many positional arguments
+  "PLR1702",  # Too many nested blocks
+  "PT012",  # pytest.raises should surround a single statement - too restrictive, author can decide
+
+  # TODO rules
+  "TD",
+  "FIX002",  # Line contains TODO, consider resolving the issue
```
stefanvanburen, line 147: I'd be a fan of trying to resolve these and remove these ignores shortly, or at least keep them inline. Feels like we've reached the point of not needing many more `TODO`s?

src/connectrpc/_server_sync.py
```diff
@@ -238,7 +240,7 @@ def __call__(
                 environ, start_response, send_trailers, protocol, headers, endpoint, ctx
             )
 
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001
```
stefanvanburen: nit: It'd be nice if we added a "reason" comment after suppressions so we know what they're suppressing and why we're doing it in each case?

src/connectrpc/code.py
```diff
@@ -1,3 +1,5 @@
+"""The Connect RPC error codes."""
```
stefanvanburen: nit: some of these module comments are just `Connect` and others are `Connect RPC`; none of them necessarily read wrongly to me, but just confirming we're ok with them being different?

connectrpc-otel/connectrpc_otel/_interceptor.py
```diff
@@ -57,14 +57,17 @@ def __init__(
...
 
         Args:
             propagator: The OpenTelemetry TextMapPropagator to use. If not
                 provided, the global default will be used.
             tracer_provider: The OpenTelemetry TracerProvider to use. If not
                 provided, the global default will be used.
+            meter_provider: The OpenTelemetry MeterProvider to use. If not
+                provided, the global default will be used.
             client: Whether this interceptor is for a client or server.
+
```
stefanvanburen, line 70: somewhat surprised we only have two of these cases where we need to add an empty line 🤷. all good, looks fine.

stefanvanburen (approved):
sweet, thanks for doing this. I basically just had a bunch of nits, none of which are blocking.

src/connectrpc/_protocol.py
anuraaga: Yeah will separately look at migrating from noqa

pyproject.toml
anuraaga: Yeah will check them out separately

PR #317, opened 2026-08-12
Use new protobuf-py plugin helpers

anuraaga:


protoc-gen-connectrpc/protoc_gen_connectrpc/__init__.py
```diff
@@ -119,7 +140,7 @@ def _generate_file(f: File, desc: DescFile, options: Options) -> None:
 
 
 def _generate_async_stubs(f: File, service: DescService, options: Options) -> None:
-    service_name = escape_identifier(service.name)
+    service_name = escape_public_identifier(service.name, python_keywords())
```
anuraaga, line 143: This should have been public identifier from before, i.e. if the proto file defines `_FooService` (almost never does), it's escaped to `pb__FooService` so it is public in Python terms.

conformance/test/gen/connectrpc/conformance/v1/service_connect.py
```diff
@@ -151,7 +151,6 @@ def bidi_stream(self, request: AsyncIterator[BidiStreamRequest], ctx: RequestCon
 
         - if the response_delay_ms duration is specified, the server should wait that
           long in between sending each response message.
-
```
anuraaga, line 154: The helper does `rstrip` instead of only `removesuffix('\n')` which is better

stefanvanburen (approved):

PR #328, opened 2026-09-02
Execute metadata interceptors at the start of the chain before parsing request

anuraaga:
Even though metadata interceptors don't need the request payload, they are always executed after parsing it. While we recommend using ASGI middleware for i.e. auth, still it's not that hard for us to also allow leading metadata interceptors to run before parse and can make sure an interceptor can block a bad / malicious request before parsing it if a user chooses to do it in an interceptor.

### DIFF
--- a/src/connectrpc/_interceptor_async.py
+++ b/src/connectrpc/_interceptor_async.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Generic, Protocol, TypeVar, runtime_checkable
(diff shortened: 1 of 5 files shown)


stefanvanburen (approved):
seems ok - stepping back though, special-casing the metadata interceptors at the front of the list for servers feels not great from a discoverability perspective. Any other Python APIs act similarly to this? anything we ought to learn from [connect-go's v2 design](https://github.com/connectrpc/connect-go/pull/951) that we ought to change here before we get to v1?

Issue #330, opened 2026-09-04
Add check that release PR bumps both published packages


anuraaga:
We have `uv run poe bump` to bump both packages together, but it seems I accidentally used `uv version --bump`. We should add a CI check for this since it's easy to mix them up.

PR #331, opened 2026-09-04
Prepare for 0.12.1

anuraaga:
0.12.0 incorrectly missed publishing the codegen package. Filed #330 to improve it soon


.gitattributes
```diff
@@ -2,3 +2,4 @@
 
 **/gen/**/*.py inguist-generated=true
 **/_gen/**/*.py linguist-generated=true
+*_connect.py linguist-generated=true
```
stefanvanburen, line 5: I thought we recently dropped this out as it was matching a real file in the repo? (I guess ideally we'd have all the gen files under `gen/`?)

stefanvanburen (approved):

.gitattributes
anuraaga: Yeah I guess moving some files will probably help with this, will try to fix it

PR #334, opened 2026-09-07
Handle services without methods in proto plugin

anuraaga:
Similar to https://github.com/bufbuild/protoc-gen-grpc-py/pull/31


connectrpc-grpcreflect/test/connectrpc/example/empty_pb.py
stefanvanburen, line 1: can add `_pb.py` and `_pb2.pyi` to the generated file list 😄

stefanvanburen (approved):

protoc-gen-connectrpc/protoc_gen_connectrpc/__init__.py
```diff
@@ -324,7 +327,7 @@ def _generate_async_stubs(f: File, service: DescService, options: Options) -> No
                     if _supports_get(method):
                         f.print("use_get=use_get,")
                 f.print(")")
-            f.print()
+        f.print()
```
stefanvanburen, line 330: actually - do we want this? looks like it's causing the diff in conformance/test/gen/connectrpc/conformance/v1/service_connect.py above, dropping the newline between methods?

stefanvanburen (changes requested):
actually, one thing claude noticed that I think we ought to fix? repro'd below (w/ nits added in case you wanted to take a pass as well)

* * *

## 1. Empty services hit a latent bug in the async server (blocking)

`src/connectrpc/_server_async.py:173` checks truthiness, not `None`:

```python
if not self._resolved_endpoints:
```

For a service with no methods `_resolve_endpoints` returns `{}`, which is falsy, so the lazy-resolve branch runs on *every* request even after lifespan startup already resolved it. When the service was passed as an async generator, that branch raises. Reproduced against `main` with an empty-endpoints ASGI app:

```
lifespan messages: [{'type': 'lifespan.startup.complete'}]
resolved endpoints after startup: {}
RuntimeError: ASGI server does not support lifespan but async generator passed for service. Enable lifespan support.
```

So `NoMethodsASGIApplication(service_gen())` mounted under a real ASGI server raises on the first request, despite lifespan having succeeded. Non-generator services don't crash but re-resolve endpoints per request.

Fix:

```python
if self._resolved_endpoints is None:
```

`test/test_lifespan.py` is the natural home for a regression test using the new `empty_connect` module. The sync/WSGI path resolves eagerly and is unaffected.

## Nits, non-blocking

- `test/proto/connectrpc/example/empty.proto:5` — `service NoMethods {}` has no leading comment, so the docstring interaction is unexercised. A documented empty service emits a docstring *and* `pass` (lines 150-151, 340-341) — valid, just redundant. Adding a comment to the test proto locks that output in.
- No golden coverage for `io=sync` / `io=async` with an empty service; no `buf.gen.yaml` anywhere sets `io=`, so that's pre-existing.
- Every generated file in the repo has exactly one service, so the inter-service separator this change touches stays untested. Notably `_generate_sync_stubs` still emits no trailing blank at all, meaning service N+1's `class` would butt directly against service N's `ClientSync`.

stefanvanburen (approved):

